PR #10, opened 2020-06-27
Fix ETC_RGB4 export

squaresmile:
Python has a special syntax for [tuple with one item](https://docs.python.org/3/tutorial/datastructures.html#tuples-and-sequences).


K0lb3:
good catch,

PR #12, opened 2020-08-09
Version as tuple for easier reading

UserUnknownFactor:
One more resource search variant
Small header read error
Some additional info in binary reader exceptions

### DIFF
--- a/UnityPy/classes/AnimationClip.py
+++ b/UnityPy/classes/AnimationClip.py
@@ -20,7 +20,7 @@ class Keyframe:
         self.value = readerFunc()
         self.inSlope = readerFunc()
         self.outSlope = readerFunc()
-        if reader.version[0] >= 2018:  # 2018 and up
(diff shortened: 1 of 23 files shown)


K0lb3:
Thanks for the great pr.
I didn't know that tuples could be used that way.
It's definitely way more readable that way.
(Previously I simply used regex to pre-convert all AssetStudio code files to something pythonic.)

Before I merge the PR as it is,
I would like to know why you want to use copy in the texture converter.
`image_data = copy(texture_2d.image_data)`

UserUnknownFactor:
I tried to access `texture_2d.image_data` after calling the converter and it was reset somewhere. This was a stop-gap measure to keep it as is. Because I needed to use `image_data` afterwards.

K0lb3:
I see,
that's weird.
Well, since that measure doesn't have any real downsides,
it should be fine.

PR #22, opened 2020-10-23
Better AssetBatchConverter

UserUnknownFactor:
There are fixes allowing reserialization of `MonoBehaviour`, `Texture2D` and `Sprite`. And updated `AssetBatchConverter`, also moved part of it to `Environment` as it's repeated in extraction/repacking. 
The stuff removed from `AssetsManager` is inconsistent with `Environment`, please check it (it doesn't return/save refs to resources inside the asset).

### DIFF
--- a/AssetBatchConverter.py
+++ b/AssetBatchConverter.py
@@ -1,109 +1,90 @@
-import os
+import os, sys
+from glob import glob
 from UnityPy import AssetsManager
 from collections import Counter
 import zipfile
+from tqdm import tqdm
 
-TYPES = ['Sprite', 'Texture2D', 'TextAsset']
+TYPES = ['Sprite', 'Texture2D', 'TextAsset', 'MonoBehaviour']
 
-ROOT = os.path.dirname(os.path.realpath(__file__))
-
-# source folder
-ASSETS = os.path.join(ROOT, 'assets')
-# destination folder
-DST = os.path.join(ROOT, 'extracted')
-# number of dirs to ignore
-# e.g. IGNOR_DIR_COUNT = 2 will reduce
-# 'assets/assetbundles/images/story_picture/small/15.png'
-# to
-# 'images/story_picture/small/15.png'
-
-IGNOR_DIR_COUNT = 2
-
-os.makedirs(DST, exist_ok=True)
-
+ROOT = os.path.abspath(os.getcwd()) # base directory
+DST = os.path.join(ROOT, "output") # destination folder
+ASSETS = os.path.join(ROOT,"input\\*.*") # source folder or file
 
 def main():
-	for root, dirs, files in os.walk(ASSETS, topdown=False):
-		if '.git' in root:
-			continue
-		for f in files:
-			print(f)
-			extension = os.path.splitext(f)[1]
-			src = os.path.realpath(os.path.join(root, f))
+    os.makedirs(DST, exist_ok=True)
+    for file_name in glob(ASSETS):
+        extension = os.path.splitext(file_name)[1]
+        src = os.path.realpath(os.path.join(ROOT, file_name))
 
-			if extension == ".zip":
-				archive = zipfile.ZipFile(src, 'r')
-				for zf in archive.namelist():
-					extract_assets(archive.open(zf))
-			else:
-				extract_assets(src)
+        am = None
+        if extension == ".zip":
+            archive = zipfile.ZipFile(src, 'r')
+            for zf in archive.namelist():
+                am = AssetsManager(archive.open(zf))
+                print("Parsing file:", zf)
+        else:
+            am = AssetsManager(src)
+            print("Parsing file:", src)
+        if am is None:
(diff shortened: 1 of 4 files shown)


K0lb3:
Thanks for the effort.
I have a bunch of unpublished commits that are in conflict with your PR,
so I will have to go through everything in detail.

K0lb3:
```python
class TextAsset(NamedObject):
    def __init__(self, reader: EndianBinaryReader):
        super().__init__(reader=reader)
        self.script = reader.read_aligned_string()
```
albeit rate, but occasionally some games put raw data into TextAssets,
so won't do

UserUnknownFactor:
Reverted the `TextAsset` stuff, thanks.

UserUnknownFactor:
I can merge the new stuff by myself when you publish it too if it doesn't conflict too much, or should I close the commit?

K0lb3:
__**EndianBinaryReader**__
The debug const should either be disabled by default, changed to a kwarg, or used as package wide variable.
It might make sense to make a new module that holds all UnityPy internal DEBUG flags.
That way they would be easily accessible.

__**AssetsManager**__
The AssetsManager is just around for backward compatibility.
The removal of ``load_file`` and ``load_zip_file``break the manager,
so that change has to be reverted.

__**AssetBatchConverter**__
The AssetBatchConverter is used as a showcase and example of how the container path can be used to restore the Unity internal structure of a project.
Your version only extracts all objects into a single folder, ignoring that so far unique aspect of UnityPy.

Overall the AssetBatchConverter is just an example, so both versions can be moved into an example folder.
I already have another script that should be listed as an example of creating a Bundle from dumped raw objects that would be added as well.

__**envrionment**__
I kept the processing part out of the environment on purpose because it's usually tailor-made for each project.
A general access point that contains all default export functionalities would make sense tho.
This access point should be easily extendable, so that custom object type handles can be added to it.

Ins

process so far kept out of it on purpose
could generate a generator tho


compression -> packer ?
if, then also changed in File prototype
-> changes and results tested against other Unity extraction tools and games?

BundleFile:
    file flag removal?

is_changed missing for WebFile

-> expansion for new assets and files

K0lb3:
__**EndianBinaryReader**__
The debug const should either be disabled by default, changed to a kwarg, or used as package wide variable.
It might make sense to make a new module that holds all UnityPy internal DEBUG flags.
That way they would be easily accessible.

__**AssetsManager**__
The AssetsManager is just around for backward compatibility.
The removal of ``load_file`` and ``load_zip_file``break the manager,
so that change has to be reverted.

__**AssetBatchConverter**__
The AssetBatchConverter is used as a showcase and example of how the container path can be used to restore the Unity internal structure of a project.
Your version only extracts all objects into a single folder, ignoring that so far unique aspect of UnityPy.

Overall the AssetBatchConverter is just an example, so both versions can be moved into an example folder.
I already have another script that should be listed as an example of creating a Bundle from dumped raw objects that would be added as well.

K0lb3:
__**envrionment**__
I kept the processing part out of the environment on purpose because it's usually tailor-made for each project.
I agree that an inbuild handler for default types makes sense,
I would design it in a different way tho.

I would add a generator to the environment that returns (container_path, obj, obj_handler, extension).
The obj handler would be selected from a list of registered handlers.
That way it would be easy to add new object class handlers, overwrite or even delete existing ones.

__**Save**__
First of all gj, but I still have to ask about some points.

The name of the compression var is defined in the File prototype that is used by WebFile, BundleFile, and SerializedFile,
so it has to be renamed there as too.
The is_changed flag should be added there as well.
WebFile should use the flag too.

Why did you remove the file flag in the BundleFile?

UserUnknownFactor:
**EndianBinaryReader**: I set the option for `False` in the code for now so please change it to the corresponding variable in the separate module when you decide to create that. Maybe move both logging and debug switches there altogether too?

**AssetsManager**: I see, reverted.

**AssetBatchConverter**: 

> container path can be used to restore the Unity internal structure of a project.

I don't understand. You mean that it doesn't creates `objfmt` folder inside `local_path`? If the code need that directory structure it can be set to`am.ignore_dir_lvls = 2` + `make_path(DST, local_path, `~~objfmt,~~` fname)` to be exactly the same as original example, no?

**envrionment**: Agree, that would be better if there was a generator. What do you want the `extension` parameter for by the way?

**Save**: I renamed the compression var already. :) The problem is that WebFile uses entirely different compressions than other formats, maybe instead of `if "supported packer" == "packer from the argument"` there should be `if "supported packer" in "packer list from the argument":`? 

> Why did you remove the file flag in the BundleFile?

Is it a file flag or block flag? I couldn't find what it corresponded to and `f.flag` throws an error.

UserUnknownFactor:
I separated the PRs for better menageability.

UserUnknownFactor:
@K0lb3 In short I don't get how you really want it. But please implement something to this extent.

PR #28, opened 2020-12-25
Make PPTr work for bare assets by keeping cache

UserUnknownFactor:
Is it okay like this, even if temporary? We can then open non-bundled assets that reference other files via PPtr this way. 
Although it's not a perfect solution due to using global cache variable.


K0lb3:
That should work for now.
I've been working on the internal file structure a bit lately,
so some changes will be made to that in a week.

PR #31, opened 2021-02-12
Fix Renderer ReadByte typo

Extacy:


Extacy:
Assuming this is a typo, should be read_byte as ReadByte doesn't exist in ObjectReader

K0lb3:
Thanks for finding and fixing that typo.

PR #33, opened 2021-02-22
Correct params order

StarHeartHunt:


K0lb3:
Good catch,
but strictly speaking, the whole merge split assets functionality has to be rewritten - or to be implemented for real once.
Atm it has a bunch of problems.

PR #36, opened 2021-03-09
fix mesh export error for unity2017.4

jesliwang:
1. Change indent_style to space。  All project files indent with space, .editorconfig's indent_style make it hard for edit
2. Change 'except:' to 'except Exception as e: \n print(traceback.format_exc())'.  With 'Except:', it's hard to decide the reason of failure
3. Update 'AnimationClip.py' and 'Mesh.py' Refer to ’Perfare/AssetStudio‘


UnityPy/classes/AnimationClip.py
```diff
@@ -80,12 +80,12 @@ def UnpackFloats(
         indexPos: int = bitPos // 8
         bitPos %= 8
 
-        scale: float = 1.0 / self.m_Range
+        scale: float = 1.0 / self.m_Range if self.m_Range != 0 else 0.0001
```
K0lb3, line 83: C#'s division by 0 results in infinity.
This isn't correctly reflected in your version.
```scale = (1.0 / self.m_Range) if self.m_Range else float("inf")``` would be better.

File: UnityPy/classes/AnimationClip.py [hunk 2]
```diff
@@ -101,7 +101,7 @@ def UnpackFloats(
                         bitPos = 0
 
                 x &= uint((1 << self.m_BitSize) - 1)  # (uint)(1 << m_BitSize) - 1u
-                data.append(x / (scale * ((1 << self.m_BitSize) - 1)) + self.m_Start)
+                data.append(x / (scale * ((1 << self.m_BitSize) - 1) if self.m_BitSize != 0 else 1 ) + self.m_Start)
```
K0lb3, line 104: Same as above - float("inf").

UnityPy/classes/AnimationClip.py
```diff
@@ -136,7 +136,7 @@ def UnpackInts(self):
         for i in range(self.m_NumItems):
             bits = 0
             entry = 0
-            while bits << m_BitSize:
+            while bits < m_BitSize:
```
K0lb3, line 139: Have you investigated this?

AssetStudio also uses <.
[AnimationClip - line 112](https://github.com/Perfare/AssetStudio/blob/dfb74baf79cf6edd746c1e4f976e3bb03ff88876/AssetStudio/Classes/AnimationClip.cs#L112)

UnityPy/classes/Mesh.py
```diff
@@ -555,57 +558,57 @@ def ProcessData(self):
...
         version = self.version
         m_VertexData = self.m_VertexData
         m_VertexCount = self.m_VertexCount = m_VertexData.m_VertexCount
-        
+
         for chn, m_Channel in enumerate(m_VertexData.m_Channels):
             if m_Channel.dimension > 0:
                 m_Stream = m_VertexData.m_Streams[m_Channel.stream]
                 channelMask = bin(m_Stream.channelMask)[::-1]
-                if channelMask[chn] == "1":
```
K0lb3, line 577:
```csharp
                var channelMask = new BitArray(new[] { (int)m_Stream.channelMask });
                byte offset = 0;
                for (int i = 0; i < 6; i++)
                {
                    if (channelMask.Get(i))
 ```
The channelMask in AssetStudio "gets" a boolean value.
Since BitArray isn't natively available in python bin is used to convert the mask into bits.

The original .get(i) wouldn't fulfill the requirement if a 0 was returned,
but in your variant, a 0 would do, as "0"  is True in python.

File: UnityPy/classes/Mesh.py [hunk 5]
```diff
@@ -666,31 +669,42 @@ def ReadVertexData(self):
...
+                    texCoordBits = m_UVInfo >> (uv * kInfoBitsPerUV)
+                    texCoordBits &= (1 << kInfoBitsPerUV) - 1
+                    if (texCoordBits & kUVChannelExists) != 0:
+                        uvDim = 1 + int(texCoordBits & kUVDimensionMask)
+                        m_UV = m_CompressedMesh.m_UV.UnpackFloats(uvDim, uvDim * 4, uvSrcOffset, self.m_VertexCount)
+                        self.SetUV(uv, m_UV)
+            else:
+                self.m_UV0 = m_CompressedMesh.m_UV.UnpackFloats(
+                    2, 2*4, 0, m_VertexCount)
             if m_CompressedMesh.m_UV.m_NumItems >= m_VertexCount * 4:  #
```
K0lb3, line 703: also belongs to the else previous branch,
-> indent

K0lb3 (changes requested):
Thanks for the PR.
I still have some comments about some parts, but it looks good overall.

1. 👍 
2. I've kept it without traceback because the normal user likely won't investigate it deeper and would simply open a new issue if it's important, but it's probably better with traceback so that the issue would contain the traceback as well.
3. in the code comments

UnityPy/classes/AnimationClip.py
jesliwang: Infinity is a better choice. I'll fix it

UnityPy/classes/AnimationClip.py
jesliwang: Infinity is a better choice. I'll fix it

UnityPy/classes/AnimationClip.py
jesliwang: Misread it ？ This fix is from << to <

UnityPy/classes/Mesh.py
jesliwang: Thank you so much for correcting the mistake！！！ I  I confused bin with byte

UnityPy/classes/AnimationClip.py
K0lb3: Oh, sorry, you're right.

K0lb3 (approved):
Looks good now.

Thanks again for the hard work.

PR #38, opened 2021-03-22
Add Shader Export

jesliwang:
make two changes

1.  split some enums from Shader.py into files for usage in  exporters
2. add ShaderConverter.py


K0lb3:
Good work overall.

There were still some small mistakes, but they were easy to fix.
I will merge the PR into the master branch after I added some tests for the converter.

K0lb3:
Thanks again for your work,
I hope that you keep it up 😏

PR #54, opened 2021-06-16
Correct a function and some indents in AnimationClip

Therhokar:
Correct StreamedClip.ReadData
Correct indents before AnimationClipBindingConstant.FindBinding


K0lb3:
Thanks, and sorry for the delay.

PR #62, opened 2021-07-04
Close function to close the opened file and use it

w3bOs:
Avoid leaving open file handles.

### DIFF
--- a/UnityPy/environment.py
+++ b/UnityPy/environment.py
@@ -16,6 +16,7 @@ class Environment:
     def __init__(self, *args):
         self.files = {}
         self.path = "."
+        self.opened = None
 
         if args:
             for arg in args:
@@ -53,8 +54,12 @@ class Environment:
     def load(self, files: list):
         """Loads all files into the AssetsManager."""
         for f in files:
+            self.opened = open(f, 'rb')
             self.files[f[len(self.path):].lstrip(
-                "/\\")] = self.load_file(open(f, "rb"), self)
+                "/\\")] = self.load_file(self.opened, self)
+
+    def close(self):
+        self.opened.close()
 
     def load_file(self, stream, parent=None):
         if not parent:


UnityPy/environment.py
```diff
@@ -53,8 +54,12 @@ def load_folder(self, path: str):
     def load(self, files: list):
         """Loads all files into the AssetsManager."""
         for f in files:
+            self.opened = open(f, 'rb')
             self.files[f[len(self.path):].lstrip(
-                "/\\")] = self.load_file(open(f, "rb"), self)
+                "/\\")] = self.load_file(self.opened, self)
```
sbrinkerhoff, line 59: I think I had recommended you PR this with a context manager.  Storing the handle into a class attribute doesn't really add value here. 

I would suggest updating this PR with the context manager example..

```
def load(self, files: list):
        """Loads all files into the AssetsManager."""
        for f in files:
            with open(f, ''rb') as handle:
                self.files[f[len(self.path):].lstrip(
                    "/\\")] = self.load_file(handle, self)
```

And drop the class attribute and `close` method.  This will ensure EVERY file is closed, and not just the case you had mentioned where you wanted to close a SINGLE flie.

w3bOs: Yeah but the problem is that the lib needs to have the file opened to work correctly.

K0lb3: > Yeah but the problem is that the lib needs to have the file opened to work correctly.

That's indeed the case.


The problem of the implementation is,
that only ONE file handle is stored that way, so if one loads multiple files,
this solution won't really do anything, as only the last file handle is saved, while the previous ones will be overwritten.
If an array of all opened files would be kept, it would work.


The big question is tho, if all of this is worth the trouble, considering that Python resolves the unused file handles itself.
Sure, this could lead in some rare cases to file handles that are kept open while the script itself is still running, but those are likely really rare, at least, I can't remember ever running into an issue related to that.

PR #64, opened 2021-07-21
Allow image source from str or non-PIL file handle

UserUnknownFactor:
This allows assigning images by their path or opened file handle.
Also adds possibility to change the cab name from outside.

### DIFF
--- a/UnityPy/classes/Texture2D.py
+++ b/UnityPy/classes/Texture2D.py
@@ -4,6 +4,7 @@ from ..export import Texture2DConverter
 from ..helpers.ResourceReader import get_resource_data
 from ..streams import EndianBinaryWriter
 from PIL import Image
+from io import BufferedIOBase, RawIOBase, IOBase
 
 
 class Texture2D(Texture):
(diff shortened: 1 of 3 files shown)


K0lb3:
Hi, great to see you again.
Looks good overall,
but we have to talk a bit about the image assignment.

The main idea behind forcing the user to pass PIL.Image was, that later on texture compression could be added to UnityPy, which would make handling other types a bit troublesome.
My idea was to add the compression to the Texture2DConverter,
so an Image opener would have to be kept within Texture2D.image instead of within the converter.
That's also the reason the converter returns bytes and not an image,
if compression would be used, it would happen within the converter, so it would have to return bytes and couldn't return a PIL image.

Regarding the importance of compression, it's required for DXT1,5 and some ECT textures that are used in shaders. I already created a lib for that and haven't come around to add it yet.

So overall, the only change that should be done regarding the texture2d for now is adding the PIL.open to the image setter.


The cab handling you added is fine and can be accepted as it is.

K0lb3:
Looks good now.

I just noticed that I forgot to mention something else last time,
and that is the value of  m_CompleteImageSize.
It corresponds to the size of the decompressed texture, so using the length of the data isn't really correct.
In practice the image size should be calculated via the original texture format multiplied by the pixel count.

PR #65, opened 2021-07-30
Fix Attribute Error for some assets when loading mesh

AutumnSun1996:
Some Mesh object has no `m_StreamData` attribute when `ProcessData` is called. (Like the one I added in test).
Thus, `AttrubuteError` will be raised and later methods like `ReadVertexData` will not be called.

This pull requests added try...expect clause inside `ProcessData` to handle these situations correctly, with a test for Mesh exporting.


K0lb3:
Looks like the test fail due to the line ending.
Strangely when I run it locally in any of my Linux distros, the tests pass.

AutumnSun1996:
I think it's because the behavior of git on windows.
When commit text file in windows, it will convert CRLF to LF, and convert backwards when checkout.
Git in linux distros will not do this, thus the example file will ends with LF, not CRLF.

Maybe we should remove all '\r' before comparing?

PR #70, opened 2021-09-28
Keeping File object names

UserUnknownFactor:
It is useful when finding what asset file asset belonged to originally.
Also FakeNode can be imported externally in the tree dumper in case it's needed.
And added some convenience functions.


UnityPy/classes/Object.py
```diff
@@ -46,7 +46,12 @@ def dump_typetree_structure(self) -> str:
         return self.reader.dump_typetree_structure()
 
     def read_typetree(self, nodes: list = None) -> dict:
-        tree = self.reader.read_typetree(nodes)
+        tree = dict()
```
K0lb3, line 49: This line is redundant.
``tree`` is assigned in both of the two branches, so there is no need to assign it beforehand.
Side note, using ``dict()`` is slower than using ``{}``, see [No, \[\] And list\(\) Are Different In Python](https://towardsdatascience.com/no-and-list-are-different-in-python-8940530168b0)

UnityPy/files/ObjectReader.py
```diff
@@ -200,13 +200,13 @@ def dump_typetree_structure(self) -> str:
...
         self.reset()
+        tree = dict()
         if nodes:
             tree = TypeTreeHelper.read_typetree(nodes, self)
         elif getattr(self.serialized_type, "nodes", None):
-            tree = TypeTreeHelper.read_typetree(
-                self.serialized_type.nodes, self)
-        else:
-            tree = {}
+            if self.serialized_type.nodes:
```
K0lb3, line 207: I backtracked a bit and noticed that serialized_type should always have the field nodes,
so instead of
```python
 elif getattr(self.serialized_type, "nodes", None):
     if self.serialized_type.nodes:
```

```python
elif self.serialized_type.nodes:
```
should do.

UnityPy/files/SerializedFile.py
```diff
@@ -368,10 +368,10 @@ def read_type_tree_blob(self, type_tree):
 
         return string_buffer_reader.bytes
 
-    def get_writeable_cab(self, name: str = "CAB-UnityPy_Mod.resS"):
+    def get_writeable_cab(self, name: str = None):
```
K0lb3, line 371: Since ``get_writeable_cab`` can be called without assigning ``name``,
a suiting name should be found by the function itself.
The name shouldn't be one of the existing resource cabs that are supposed to be read-only.

UserUnknownFactor: Yeah, that was from my personal fork, forgot to revert.

Issue #67, opened 2021-08-08
Should avoid changing `_value` in Enum._missing_


AutumnSun1996:
**Code**
```python
from UnityPy.enums import ClassIDType

# two unkonw values
t1 = ClassIDType(-10)
t2 = ClassIDType(-20)

assert t1 == -10, (t1, t1.value)
```
**Error**
```
AssertionError: (<ClassIDType.UnknownType: -1>, -20)
```

New unkonw value will change the value of previous values.

**Bug**
The bug is caused by `Enum._missing_`

https://github.com/K0lb3/UnityPy/blob/5b0a688ad7e7cbe08897e9f901ca7dbe8dbb3bac/UnityPy/enums/ClassIDType.py#L384-L388

`ClassIDType.UnknownType` is singlenton, so changes to `_value` should be avoided.

`BuildTarget` will suffer the same issue.

Didn't find a easy method to fix this. Maybe a custom Enum implemention is reqeuired.

PR #71, opened 2021-09-30
Change how enums are created, fixes K0lb3/UnityPy#67

UserUnknownFactor:
Is it viable this way? It's better if the value stores original data if there is no corresponding enum field so we can later write it back as is.

### DIFF
--- a/UnityPy/classes/AnimationClip.py
+++ b/UnityPy/classes/AnimationClip.py
@@ -3,7 +3,7 @@ from enum import IntEnum
 
 from .NamedObject import NamedObject
 from .PPtr import PPtr
-from ..enums import ClassIDType
+from ..enums import ClassIDType, makeClassID
 from ..math import Quaternion, Vector3
 from ..streams import EndianBinaryReader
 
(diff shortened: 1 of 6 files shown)


K0lb3:
It is indeed a better solution than the current system that definitely causes problems once multiple instances with different unknown values pop up. The general idea of what I intended to do with it was to make sure that the field will definitely be of the specific enum type.

I think that your implementation would break that idea and return an int instead if the number is not within the enum class.
```python
def makeBuildTarget(_bt):
    if isinstance(_bt, BuildTarget):
        return _bt
    elif BuildTarget.has_value(_bt):
        return BuildTarget(_bt)
    return _bt
    ```
    
    An alternate solution for that part  would have been
    ```python
    class X(IntEnum):
    @classmethod
    def _missing_(cls, value):
        return value
    ```
    
I guess there are only three real solutions for the problem,
1. What you plan to do, trying to let all comparisons with instances of the enum classes be done via __eq__,
2. Extending the enum class within __missing__ e.g. via [extend_enum from aenum](https://github.com/ethanfurman/aenum/blob/731cd6e54772071779a509f2044d71e53a028d16/aenum/__init__.py#L3295) which would be quite hacky.
3. Using a (custom) type that can handle unknown values but behaves like an enum otherwise.

UserUnknownFactor:
> make sure that the field will definitely be of the specific enum type.

It can't always be strictly `Enum` unless we know all possible values beforehand. It can be only Enum-like anyway as this is meta-project that heavily depends on unknown parent's project behaviour. So it's better to have possibility of several unknown types.

PR #82, opened 2022-01-24
Shader fix for Unity 2020 and newer

deibu:
Updated Shader.py to reflect changes in AssetStudio in order to support shaders in 2020 and newer versions of Unity.


K0lb3:
Thanks a lot.

PR #89, opened 2022-02-13
Support editing of GameObject.m_IsActive true|false

ablegao:
Support editing of GameObject.m_IsActive true|false
So, it looks like it's working, but I'm not sure the way it's written is up to spec


K0lb3:
Sorry for replying to you that late as well.

Please check [my other reply for a similar issue](https://github.com/K0lb3/UnityPy/pull/86#issuecomment-1059858990) for details.

PR #90, opened 2022-02-15
Fix Big Big Big Bug

ablegao:
When merging multiple splitx files into the one file, the wrong code logic caused the disk to be full


K0lb3:
Sorry that I didn't respond so far.
Instead of creating a new file I think it's better to keep the merged file in memory than to create it locally, which is why I made an alternate solution to your pr that would have fixed the issue as well.

PR #92, opened 2022-03-01
fix shader format for Unity 2021.2

jerryhwq:
According to the AssetStudio code, Unity Shader format has changed after Unity 2021.2.


K0lb3:
Good catch, thanks.

PR #97, opened 2022-03-22
Remove unused tkinter import

Parnassius:
This caused UnityPy to crash on import if tkinter was not installed.
The import was originally added in 7a45608bfc34bee6c2267a528274fd0c961d6f39


Parnassius:
I guess it's because actions/setup-python installs the whole python stdlib, while on distros parts of it are often in some separate packages, like python3-tkinter or python3-tk for tkinter

K0lb3:
That makes sense.

PR #123, opened 2022-08-10
Fix save_fs flag confusion

astiob:
See commit messages: the argument naming is opposite from field naming, causing confusion and ruining `save("original")`. Also, `save_fs` always writes `DirectoryInfo` but doesn’t enforce the corresponding flag, making it easy to create bundles that Unity can’t load. This PR renames the arguments and adds a check that informs the user the flag is required.


K0lb3 (approved):
Thanks for this PR.
Looks like I really messed up when I wrote the code,
so thanks for noticing and fixing this issue.

PR #133, opened 2022-09-11
Add filter functionality to extractor

sparr:
This PR adds a new optional parameter to extract_assets and export_obj that will skip exporting objects based on a provided filter function.

Here is an example function from my real use case, which only exports audio clips over a certain length with names that match one of a list of file glob patterns and does not match any of another list of file glob patterns.

```python3
def asset_filter(obj):
    # Only export AudioClip assets
    if obj.type != ClassIDType.AudioClip:
        # incorrect type
        return False
    # A filter is called twice for each object
    # First, during extract_assets, when only metadata has been loaded
    # Second, during extract_obj, after the asset has been .read()
    # The checks here for m_Length and name will be skipped during the first pass
    if obj.m_Length is not None and obj.m_Length < args.minduration:
        # too short
        return False
    if obj.name is not None:
        for excludespec in assetsexcludespec:
            if excludespec and fnmatch.fnmatch(obj.name, excludespec):
                # excluded by an excludespec match
                return False
        for filespec in assetsfilespec:
            if filespec and fnmatch.fnmatch(obj.name, filespec):
                # included by a filespec match
                return True
        # not included by any filespec match
        return False
    # correct type, length and name checks skipped because they aren't available
    return True
```

The diff is a lot cleaner with whitespace changes omitted. https://github.com/K0lb3/UnityPy/pull/133/files?w=1

### DIFF
--- a/UnityPy/tools/extractor.py
+++ b/UnityPy/tools/extractor.py
@@ -18,6 +18,7 @@ from UnityPy.classes import (
 from UnityPy.enums.ClassIDType import ClassIDType
 from typing import Union, List, Dict
 from pathlib import Path
+from collections.abc import Callable
 
 
 def export_obj(
@@ -26,6 +27,7 @@ def export_obj(
     append_name: bool = False,
     append_path_id: bool = False,
     export_unknown_as_typetree: bool = False,
+    asset_filter: Callable[[Object], bool] = None,
 ) -> List[int]:
     """Exports the given object to the given filepath.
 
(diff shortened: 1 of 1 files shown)


UnityPy/tools/extractor.py
```diff
@@ -98,35 +107,38 @@ def defaulted_export_index(type: ClassIDType):
...
-            )
-            os.makedirs(os.path.dirname(obj_dest), exist_ok=True)
-            exported.extend(
-                export_obj(
-                    obj,
-                    obj_dest,
-                    append_path_id=append_path_id,
-                    export_unknown_as_typetree=export_unknown_as_typetree,
+            # The filter here can only access metadata. The same filter may produce a different result later in extract_obj after obj.read()
+            if (not asset_filter) or asset_filter(obj):
```
K0lb3, line 111: To keep the indention low,
it would be better to use
```python
if asset_filter is not None and not asset_filter(obj):
    continue
```

Issue #45, opened 2021-04-16
Sprite-Texture2D merging issue


szescxz:
**Code**
```Python
import UnityPy
env = UnityPy.load("live_note_default")
for path, obj in env.container.items():
    print(path, obj.type)
````

**Error**
N/A

**Bug**
Seems like the Sprite-Texture2D merging is performed in the wrong direction.
Four `Sprite` in the sample: `notes_crtcl`, `notes_flick`, `notes_long` and `notes_normal`, all based on the `Texture2D` `notes`. (let's ignore the rest of the items here)
I'm expecting all four `Sprite`s in the output, but UnityPy only gives me one `Sprite` named `notes`.

Also a tiny bug, too lazy for a separate issue: 1.7.3 breaks `str(obj.type)`, now all items are `UnknownType`.

**To Reproduce**
- a copy of the file that causes the problem
[live_note_default.zip](https://github.com/K0lb3/UnityPy/files/6326043/live_note_default.zip) (unzip it and you should get an asset bundle file)
- following data:
  - Python version 3.7.3
  - UnityPy version 1.7.2

**Comment by K0lb3** [comment]
A lot of sprites in that asset have the identical container path, so extracting the objects only via the container won't do in this case.
Strictly speaking UnityPy doesn't handle the containers correctly. UnityPy handles them as dict (path -> obj), while they are actually a multi-dict, which can have multiple entries with the same key.

> Also a tiny bug, too lazy for a separate issue: 1.7.3 breaks str(obj.type), now all items are UnknownType.

Thanks for reporting that issue, I pushed a fix for it just now.

**Comment by erikchan002** [comment]
Are there any plans on completing to TODO of fixing the container dictionary for non-unique container paths?

While I understand and can workaround the fact that I cannot access all of the objects via the container, the issue also prevents me from knowing the container path for some of the objects even when the objects are accessed by `env.objects`:
```python
env = UnityPy.load("live_note_default")
path_ids = [-8615770281572025746,-8198049756159249405,-3594623656469114911,310693865538495986,4976375391278443517]
[object.container for object in env.objects if object.path_id in path_ids]
"""
[None,
 'assets/sekai/assetbundle/resources/startapp/live/note/default/textures/notes.png',
 None,
 None,
 None]
"""
```
![image](https://user-images.githubusercontent.com/10881196/188259776-5d316ca3-2557-4a1f-8223-2f6ee5ae461d.png)
<br/>
[...]

PR #134, opened 2022-09-12
Create `list_container` as `container` alternative for multiple objects with the same container path

erikchan002:
Temporary workaround for #45 without breaking `container` compatibility before version 2

`SerializedFile`'s `_container` is also changed to be constructed using the `list_container` alternative so that all assets with a container path will properly show it instead of `None` when accessed with `.container`

A `multiple_objects_per_container` flag is introduced to `ResourceManger.save()` and `extract_assets()` in `extractor.py` to utilize `list_container` with a `False` default to avoid breaking compatibility

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ One of these objects can be an AssetBundle, which contains a file path for some 
 
 All objects can be found in the ``.objects`` dict - ``{ID : object}``.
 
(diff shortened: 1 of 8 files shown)


K0lb3:
implemented by https://github.com/K0lb3/UnityPy/pull/186

PR #138, opened 2022-09-24
Handle exportMonoBehavior for Object without serialized type or nodes

sparr:
While exporting assets from the game "As Far As The Eye" I encountered a call to exportMonoBehaviour for an Object (not a MonoBehaviour), when the Object failed the `if obj.serialized_type and obj.serialized_type.nodes:` test. This led to:

```
  File "/home/sparr/.local/share/virtualenvs/vgm-extractor-1oHW3IDG/src/unitypy/UnityPy/tools/extractor.py", line 219, in exportMonoBehaviour
    if not export:
UnboundLocalError: local variable 'export' referenced before assignment
```

So I've initialized it here to avoid this error.


K0lb3:
Oh, that's indeed a problem.

PR #140, opened 2022-09-24
Disambiguate path_id collision using obj type. Change extract/export return type.

sparr:
When multiple assets have the same `path_id` and `use_container` is `False` then only the first such asset will be exported. This patch stores a tuple of `(obj.type, obj.path_id)` instead of just `obj.path_id` to disambiguate and allow all the assets to pass the check.

This patch changes the return type of `extract_assets`, `export_obj`, and all of the type-specific export functions. Instead of returning `list[int]` they now return `list[tuple[ClassIDType, int]]`. This is a breaking change, although I cannot find any current uses of extract_assets or export_obj on github that would be broken.

It would be possible to `return [path_id for _,path_id in exported]` to avoid the breaking change, but that would continue the current state of producing an ambiguous list of path_ids that can't be resolved back to specific assets.

I do not (that I know of) have an assets file with which to test some of the code here, specifically the Sprite.m_RD.alphaTexture logic.

### DIFF
--- a/UnityPy/tools/extractor.py
+++ b/UnityPy/tools/extractor.py
@@ -130,7 +130,7 @@ def extract_assets(
         for obj in objects:
             if asset_filter is not None and not asset_filter(obj):
                 continue
-            if obj.path_id not in exported:
+            if (obj.type, obj.path_id) not in exported:
                 exported.extend(
                     export_obj(
                         obj,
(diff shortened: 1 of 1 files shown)


K0lb3:
I had this idea some time ago as well, but there is a better and safer option.

Using a combination of class type and path id could still result in problems in case of specific objects like the AssetBundle (always at path id 1), or some other objects that seem to start at numbers and increase instead of using "random" numbers.

My conclusion of this was, that the best bet would be to combine the path_id with the asset file the object is stored in,
so (obj.path_id, obj.assets_file).

sparr:
Thanks for the catch. This also came up on Discord. I've made the change, but with the tuple in the other order so they can be sorted lexicographically for other purposes. Sorry for the two extra commits showing, that's a rebase on top of master after recent PR merges; I'm not sure what I did wrong for them to show up here.

K0lb3:
Sorry that I forgot about it.
It looks good now, so thanks a lot :D

PR #172, opened 2023-03-11
Breaking Change: Fix objects with the same container path

t-wy:
Addressing the issue #45, referenced PR #134 and AssetStudio's original implementation.

Reimplemented the array of KeyValuePair (multi-dict) structure in AssetStudio as a list-of-tuple structure to address the issue of multiple objects with the same container path when implemented as dict, such that all objects return the expected .container str value instead of None.

Breaking Part: All .items() call to .container or .m_Container is no longer needed as it is an array instead of a dict.
To avoid the compatibility issue, a list wrapper class with an .items() method that returns the list may be used in all container properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ One of these objects can be an AssetBundle, which contains a file path for some 
 
 All objects can be found in the `.objects` dict - `{ID : object}`.
 
-The objects with a file path can be found in the `.container` dict - `{path : object}`.
+The objects with a file path can be found in the `.container` list - `[(path, object)]`.
 
 ### [Object](UnityPy/files/ObjectReader.py)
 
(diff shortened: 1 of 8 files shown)


K0lb3:
Sorry for the late reply.
I'm aware of the issue and refrained from fixing it because of the issue you mentioned - that it would be a fatal breaking change.

I "fixed" it in a test branch some time ago without breaking the normal usage pattern while still allowing it to pass through all keys and values.
[UnityPy@fsspec - UnityPy/files/SerializedFile.py](https://github.com/K0lb3/UnityPy/commit/3b3d74d53da0ea7493ef2a199a1e81e07369a549)

t-wy:
Thanks for the reply. I see the concern, especially when it comes to the compatbility issues.
Seems like the workaround wrapper provided is within my expectations. Looking forward to the release than :)
(Meanwhile I can still take some monkey patches to handle my own usages :))

K0lb3:
implemented by #186

PR #178, opened 2023-04-15
BundleFile: fix exception on file.save("original")

astiob:
Regression in commit https://github.com/K0lb3/UnityPy/commit/9232cdd57e93f140b2d7431905baff40e05488bc (release 1.9.21), which renamed `_data_flags` to `dataflags`.

### DIFF
--- a/UnityPy/files/BundleFile.py
+++ b/UnityPy/files/BundleFile.py
@@ -223,7 +223,7 @@ class BundleFile(File.File):
             elif packer == "original":
                 self.save_fs(
                     writer,
-                    data_flag=self._data_flags,
+                    data_flag=self.dataflags,
                     block_info_flag=self._block_info_flags,
                 )
             elif packer == "lz4":


astiob:
On a slightly related note (I’d open a separate GitHub issue, but it isn’t exactly a bug), `save_fs` should maybe throw if the `data_flags` argument includes the encryption flag, because (as far as I can tell) UnityPy doesn’t implement encryption upon saving. But this isn’t entirely trivial to implement because the argument may be passed as a raw integer, so you’d need to duplicate the [version-dependent logic](https://github.com/K0lb3/UnityPy/blob/7fff866e98cb24d91f86cee29d5e2f789bd41468/UnityPy/files/BundleFile.py#L95-L108) from `read_fs` just to figure out what the integer means.

K0lb3:
Thanks for noticing and fixing the issue.
And you're right, it should error in case it's supposed to store an encrypted bundle.

PR #187, opened 2023-06-06
Fix ContainerHelper from being non-iterable

bloodwiing:
It's a very simple fix, one line and probably a simple mistake, but this single line caused my whole project to break when I updated the UnityPy version haha


K0lb3:
Oh, dang, yeah, that's indeed a stupid mistake I made there...and missed seeing during two reviews -.-'

Thanks a lot.

PR #190, opened 2023-07-04
Fix random ResourceReader file loading

RealSupremium:
You may not believe it (or you probably did already), but `possible_names` was actually a dictionary type before. The problem with that is you get these random numbers as keys, and the items inside go out of order.

image


K0lb3:
It doesn't create a dict but a set.
The idea behind it is to prevent duplicates, which used to be relevant before using .lowercase for the lookup.
Now it's indeed not necessary to use a set anymore and a list would do, indeed.

PR #191, opened 2023-07-04
Handle relative and case insensitive file paths

sparr:
This PR resolves a few problems I had exporting assets from the game Autonauts.

1. One assets file contains a reference to another with a relative path, which UnityPy was trying to load from the script's CWD instead of the original asset file path.
2. That reference doesn't match the case of the actual directory or file name.
3. There is another reference to `library/unity default assets` which, in addition to being relative and case insensitive, also doesn't exist because `Library/` is apparently a magic path in Unity. Rather than calling that out explicitly here, I've opted to just silently ignore nonexistent paths/files.

I wasn't able to test on Windows, which might have a problem with the case insensitive search.

PS: I also spotted a bug in `save()` but can't test the fix I included here. Let me know if that's a problem and I can remove it.

### DIFF
--- a/UnityPy/environment.py
+++ b/UnityPy/environment.py
@@ -112,7 +112,7 @@ class Environment:
                     file = os.path.join(self.path, file)
                 # Unity paths are case insensitive, so we need to find "Resources/Foo.asset" when the record says "resources/foo.asset"
                 if not os.path.exists(file):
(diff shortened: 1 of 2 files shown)


sparr:
Lots of unrelated `pip install` errors in the test failures. Unclear if that's all of them.

PR #208, opened 2023-09-28
Fixed "Exception: String not terminated".

AhmedAhmedEG:
Fixed a bug that caused using Environment.load() on a folder that contains plain text file independent of it's extension, to cause an "Exception: String not terminated".

### DIFF
--- a/UnityPy/environment.py
+++ b/UnityPy/environment.py
@@ -274,7 +274,7 @@ class Environment:
                 data = b"".join(data)
                 path = basepath
             else:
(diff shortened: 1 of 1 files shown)


K0lb3:
Environment.load wants a list of files as input and not a folder.
I agree that it's a confusing situation considering the behavior of UnityPy.load.

AhmedAhmedEG:
I see, normally I just give it the game folder path and all the magic happens.

PR #213, opened 2023-10-15
fix: Read SpriteBone correctly

Nattsu39:
The rotation value in the SpriteBone should be read via read_quaternion. (Only tested in version 2020.3.38)

Reference: https://github.com/Perfare/AssetStudio/pull/918


K0lb3:
Thanks for the PR.
The rotation should indeed be a quaternion, and that is for all versions, so all good.

PR #222, opened 2024-01-05
Support reading of AnimationEvent from AnimationClip

ysmu:
Thanks for the great project!

Reference: https://github.com/Perfare/AssetStudio/blob/d158e864b556b5970709c2a52e47944d53aa98a2/AssetStudio/Classes/AnimationClip.cs#L1036-L1051


K0lb3:
Looks good, thanks.

PR #239, opened 2024-06-05
Fix Unity 2022.1 Support

mos9527:
[AssetStudio introduced its Unity 2022.1 update in its last commit](https://github.com/Perfare/AssetStudio/commit/d158e864b556b5970709c2a52e47944d53aa98a2), which isn't thoroughly implemented here.
This PR adds these new properties. Which addresses https://github.com/mos9527/sssekai_blender_io/issues/3


K0lb3:
Thanks :+1:

PR #243, opened 2024-06-19
Add split file reading

EoF-1141:
When loading dependencies it's possible the file is split without containing info about being split. 
This will search for split files and load them accordingly


K0lb3 (approved):

K0lb3:
👍

PR #247, opened 2024-07-25
Implement saving Unity Web and Raw bundles

0x07E5:


UnityPy/files/BundleFile.py
```diff
@@ -399,6 +396,89 @@ def save_fs(self, writer: EndianBinaryWriter, data_flag: int, block_info_flag: i
...
+            file_content_writer.write_bytes(file_data)
+            current_offset += file_size
+
+        directory_info_writer.write(b'\x00' * file_info_header_padding_size)
+        uncompressed_directory_info = directory_info_writer.bytes
+        uncompressed_file_content = file_content_writer.bytes
+
+        # Combine directory info and file content
+        uncompressed_content = uncompressed_directory_info + uncompressed_file_content
+        compressed_content = CompressionHelper.compress_lzma(uncompressed_content)
```
K0lb3, line 455: Going by your other PR, it should only compress when the signature is UnityWeb.

0x07E5:
Added.

UnityPy/files/BundleFile.py
```diff
@@ -399,6 +396,89 @@ def save_fs(self, writer: EndianBinaryWriter, data_flag: int, block_info_flag: i
...
+        directory_info_writer.write(b'\x00' * file_info_header_padding_size)
+        uncompressed_directory_info = directory_info_writer.bytes
+        uncompressed_file_content = file_content_writer.bytes
+
+        # Combine directory info and file content
+        uncompressed_content = uncompressed_directory_info + uncompressed_file_content
+        compressed_content = CompressionHelper.compress_lzma(uncompressed_content) if self.signature == "UnityWeb" else uncompressed_content
+        
+        # Write header
+        header_size = 60 # file_header 28, metadata max 4 * 8
```
K0lb3, line 458: I don't think that it's good to assume a static value,
as it depends on the version as well as the length of version_player and version_engine that can be modified,
besides if I'm missing something.

So,
from the shared header parts:

string
+4
string
string

and for web itself:
version >= 4:
	+ 16
	+ 4

4
4
4
4 - node count -> nodes
	(4 * 2 * (n-1)) (skipped nodes)
	4
	4
version >= 2
	4
version >= 3
	4
------

version < 2
16 + 8 * n
version < 3
+4 = 20 + 8*n
version < 4
+4 = 24 + 8*n
version >= 4
+20 = 44 + 8*n

K0lb3: Ah, I see, so you try to hide that via the padding, which is a good idea,
besides....if one of the earlier strings are too long due to modifications.

K0lb3 (commented):
Looks good overall,
and thanks for the work.

I will make some smaller changes at Sunday and then merge this PR.

Issue #237, opened 2024-05-30
Can't load Unity 4.7.2f1 files while AssetStudio does.


0x07E5:
**Code**
A snippet of the code section that cases the bug or error.

```
env = UnityPy.load("example.unity3d")
```

**Error**
The error message that is produced by python.
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\0x07E5\AppData\Local\Programs\Python\Python310\lib\site-packages\UnityPy\__init__.py", line 8, in load
    return Environment(*args, fs=fs, **kwargs)
  File "C:\Users\0x07E5\AppData\Local\Programs\Python\Python310\lib\site-packages\UnityPy\environment.py", line 46, in __init__
    self.load_file(arg)
  File "C:\Users\0x07E5\AppData\Local\Programs\Python\Python310\lib\site-packages\UnityPy\environment.py", line 137, in load_file
    f = ImportHelper.parse_file(
  File "C:\Users\0x07E5\AppData\Local\Programs\Python\Python310\lib\site-packages\UnityPy\helpers\ImportHelper.py", line 139, in parse_file
    f = files.BundleFile(reader, parent, name=name, is_dependency=is_dependency)
  File "C:\Users\0x07E5\AppData\Local\Programs\Python\Python310\lib\site-packages\UnityPy\files\BundleFile.py", line 40, in __init__
    m_DirectoryInfo, blocksReader = self.read_web_raw(reader)
  File "C:\Users\0x07E5\AppData\Local\Programs\Python\Python310\lib\site-packages\UnityPy\files\BundleFile.py", line 72, in read_web_raw
    uncompressedBytes = CompressionHelper.decompress_lzma(
  File "C:\Users\0x07E5\AppData\Local\Programs\Python\Python310\lib\site-packages\UnityPy\helpers\CompressionHelper.py", line 39, in decompress_lzma
    return dec.decompress(data[5:])
_lzma.LZMAError: Corrupt input data
```

**Bug**
A description of what you expect to happen/see and what actually happens.

Can't load Unity 4.7.2f1 files.

**To Reproduce**
- a copy of the file that causes the problem
[examples.zip](https://github.com/K0lb3/UnityPy/files/15505838/examples.zip)
Inside the package:
  - example-android.unity3d Android sample, AssetStudio can load.
  - example-ios.unity3d iOS sample, AssetStudio can load.
[...]

PR #250, opened 2024-08-18
Fix LZMA decompression

0x07E5:
In Unity LZMA file, headers [5:13] is decompressed size, should be skipped.
This PR solves the decompression issue #237

### DIFF
--- a/UnityPy/helpers/CompressionHelper.py
+++ b/UnityPy/helpers/CompressionHelper.py
@@ -36,7 +36,7 @@ def decompress_lzma(data: bytes) -> bytes:
             }
         ],
     )
-    return dec.decompress(data[5:])
+    return dec.decompress(data[13:])
 
 
 def compress_lzma(data: bytes) -> bytes:


K0lb3:
You're right,
going by the [lzma specs](https://github.com/jljusten/LZMA-SDK/blob/781863cdf592da3e97420f50de5dac056ad352a5/DOC/lzma-specification.txt#L23) it should be [13:].

I guess that fixing this for good requires looking a bit into the lzma code of the python package.

AXiX-official:
In LZMA1, headers [5:13] is decompressed size, that's right
but in unity, decompressed  wouldnt be write into compressed data

K0lb3:
Hm, I guess one simply solution to the issue would be simply checking if 8:13 contains the decompressed length.
If it does, start at 13, if not, start at 8.
This would have to be stored somewhere for saving afterwards.

AXiX-official:
Sorry I may not have a complete understanding of the problem before commenting
Did you encounter a file with decompressed length saved in the lzma data that failed to read?
My understanding of this part is basically from AssetStudio and AssetsTools.NET, at least for BundleFile, AssetsTools.NET encodes as LZMA1(in cpp use func lzma_alone_decoder) and indeed does not keep the decompressed length in the compressed data.
And the format of the lzma compressed ab file I exported using Unity is also like this
I don't know much about the rest, is there a sample file that causes the problem?

PR #263, opened 2024-09-11
Implement chunk-based lz4 compression

AXiX-official:
According to [Unity's documentation](https://docs.unity3d.com/Manual/AssetBundles-Cache.html) lz4(hc) compressed AssetBundle should be chunk-based

### DIFF
--- a/UnityPy/files/BundleFile.py
+++ b/UnityPy/files/BundleFile.py
@@ -301,31 +301,23 @@ class BundleFile(File.File):
 
         file_data = data_writer.bytes
         data_writer.dispose()
-        uncompressed_data_size = len(file_data)
 
-        # compress the data
-        switch = block_info_flag & 0x3F
-        if switch == 1:  # LZMA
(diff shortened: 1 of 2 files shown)


UnityPy/files/BundleFile.py
```diff
@@ -301,31 +301,23 @@ def save_fs(self, writer: EndianBinaryWriter, data_flag: int, block_info_flag: i
...
-        # block count
-        block_writer.write_int(1)
-        # uncompressed size
-        block_writer.write_u_int(uncompressed_data_size)
-        # compressed size
-        block_writer.write_u_int(compressed_data_size)
-        # flag
-        block_writer.write_u_short(block_info_flag)
+        block_writer.write_int(len(block_info))
+        for u_size, c_size, f in block_info:
```
K0lb3, line 314: Can you please use the complete meaning as name,
so block_uncompressed_size, block_compressed_size, block_flag.

K0lb3: Thanks.

PR #264, opened 2024-09-12
1.20 - replaced hard-coded with generated classes

K0lb3:
- generate classes dynamically based on tpk
- use typetree reader to load and store all objects
- make handlers compatible

TODO:
- [x] update C typetree loader
- [x] check nodes input, convert list to hierarchy on the fly


K0lb3 (commented):
/

PR #279, opened 2024-10-25
boost ArchiveStorageDecryptor.decrypt_block with cpp implementation

AXiX-official:
On my device loading UnityCN encrypted file via
`env = UnityPy.load(“test”)`
For the original Python implementation, a 10MB file takes roughly 7.8s, and a 50MB file takes roughly 30s
But the cpp implementation only takes 0.076s and 0.24s respectively.
I can provide some test files if needed
(Only tested it on windows using msvc compilation, but I guess other platforms won't have any problems with it)


UnityPy/helpers/ArchiveStorageManager.py
```diff
@@ -101,6 +106,10 @@ def __init__(self, reader: EndianBinaryReader) -> None:
         )
 
     def decrypt_block(self, data: bytes, index: int):
+
+        if UnityPyBoost:
+            return UnityPyBoost.decrypt_block(bytes(self.index), bytes(self.substitute), data, index)
```
K0lb3, line 111: The bytes cast for self.index and self.substitute is redundant as they are already bytes objects.
Given the buffer protocol it wouldn't even be an issue if they would be bytearrays

UnityPyBoost/ArchiveStorageDecryptor.cpp
```diff
@@ -0,0 +1,92 @@
...
+    return offset;
+}
+
+PyObject *decrypt_block(PyObject *self, PyObject *args) {
+    PyObject *py_index_bytes;
+    PyObject *py_substitute_bytes;
+    PyObject *py_data;
+    uint64_t index;
+
+    if (!PyArg_ParseTuple(args, "OOOi", &py_index_bytes, &py_substitute_bytes, &py_data, &index)) {
```
K0lb3, line 61: I guess that you might have copied this approach from my code.
Instead of using O one can also directly let python fill a Py_buffer object via ``y#``.
Mind that even then the buffer has to be freed afterwards. This entails a check and possibly free if the parsetuple function fails and would immediately return.

UnityPyBoost/ArchiveStorageDecryptor.cpp
```diff
@@ -0,0 +1,92 @@
...
+        return NULL;
+    }
+
+    PyObject *result = PyBytes_FromObject(py_data);
```
K0lb3, line 65: Please create a new buffer instead of basically trying to re-using the input data.
So 
``PyObject* result = PyBytes_FromStringAndSize(null, py_data_view.len);``
followed by
``char* result_raw = PyBytes_AS_STRING(result);``
to get the raw underlying data of the result object.

``PyBytes_AS_STRING`` is allowed to be used on newly created bytes object within C if they were created with a null ptr as above.

UnityPyBoost/ArchiveStorageDecryptor.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+    Py_buffer view;
+    if (PyObject_GetBuffer(result, &view, PyBUF_SIMPLE) != 0) {
```
K0lb3, line 68: This would therefore become redundant.

UnityPyBoost/ArchiveStorageDecryptor.cpp
```diff
@@ -0,0 +1,92 @@
...
+        return NULL;
+    }
+
+    if (!PyBytes_Check(py_index_bytes) || !PyBytes_Check(py_substitute_bytes)) {
```
K0lb3, line 72: Same for this given the first comment.

UnityPyBoost/ArchiveStorageDecryptor.cpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    return offset;
+}
+
+PyObject *decrypt_block(PyObject *self, PyObject *args) {
+    PyObject *py_index_bytes;
+    PyObject *py_substitute_bytes;
+    PyObject *py_data;
+    uint64_t index;
```
K0lb3, line 59: As the index is an index to the data from python it's better to use [Py_ssize_t](https://docs.python.org/3/c-api/intro.html#c.Py_ssize_t).
This is also the type that is used by the buffer protocol for the length, making the cast below unnecessary.

UnityPyBoost/ArchiveStorageDecryptor.cpp
K0lb3: This would then also change the return type within the pyi to bytes.

K0lb3 (changes requested):
The core part looks good.
I just want some changes to the Python api usage.

AXiX-official:
In fact, while `self.index` has a type hint of bytes but is assigned as a bytearray
guess u mean `y*` instead of `y#`  [refer](https://docs.python.org/3.10/c-api/arg.html#strings-and-buffers)
If `self.index` is acctually a bytes obj then maybe I can use `y#` so I don't have to deal with buffer and reference counting I guess
However, the official documentation says that `y*`  **is the recommended way to accept binary data**
So I'm not really sure which to use.

K0lb3 (approved):

K0lb3:
You're right, I meant ``y*`` and not ``y#``.

Thanks for the adjustments and also again for the work.

PR #281, opened 2024-11-08
feat: arm64 for fmod and AudioClipConverter

martinwang2002:
The issue with arm64 Linux and FMOD is that the default audio output is in WAV PCM float format. When attempting to write the binary file in WAV PCM 16 format, the audio becomes distorted.


K0lb3:
@martinwang2002 from which fmod version did you take the arm64 fmod lib?

K0lb3:
ok, thanks

PR #285, opened 2024-11-14
Fix: some type hints

Nattsu39:
Not sure if there is a better way to implement type hints for legacy patches...


UnityPy/classes/generated.py
K0lb3, line 1: The generated file is generated by a script and isn't to be modified.
Instead of modifying it, a different approach can be taken -  creating an ``__init__.pyi`` in ``UnityPy/classes``, where the related adjustments can be made.

K0lb3 (changes requested):

UnityPy/classes/generated.py
Nattsu39: > The generated file is generated by a script and isn't to be modified. Instead of modifying it, a different approach can be taken - creating an `__init__.pyi` in `UnityPy/classes`, where the related adjustments can be made.

Type stubs don't seem to work with class/type based overrides, which means that all `generated` classes need to be defined in `__init__.pyi`. Is this acceptable?

K0lb3: That's not necessary.
I just looked into it myself, and it seems to be enough to create a .pyi for legacy patch with the adjusted stubs.

K0lb3: That said, the long term solution I have in mind is a bit different than the current solution.

Instead of patching the classes I want to move to handler classes, which basically accomplish the same as the patch.
The handler classes will get the object as input, and then handle all the post-processing, as well as export and import.

This solves the typing issue, and makes the code and usage more clear.

Nattsu39: After more attempts, I figured it out.
For mypy, creating a `.pyi` for the legacy patch works, but doing so under pyright still gives an error (I've been using pyright).
I'm trying to figure out a way...

K0lb3: Thanks for looking into this.

One attempt that worked for me with Pyright is importing the original generated class via as under a different name,
then inheriting from it and applying the patches in the inherited class.
But....that's not really a great solution in a .py itself, and also a bit confusing when one tries to lookup fields,
as it then shows the inheriting class and not the original, which might just be fine if all fields are copied.

K0lb3 (commented):
Looks pretty good now, and thanks for figuring this out.
I still have something to complain about 😛 

... / pass isn't necessary if the class has declared fields,
so all of them can be removed from the classes.

The .py files use __all__ to restrict the export to the class itself,
so the ``_`` helper function can't be imported.
Instead of defining them as separate entities, you can directly put them into the class stubs themselves.

e.g. for AudioClip

```py
from typing import Dict, List, Optional

from UnityPy.classes.generated import SampleClip, StreamedResource

class AudioClip(SampleClip):
  m_Name: str
  m_3D: Optional[bool] = None
  m_Ambisonic: Optional[bool] = None
  m_AudioData: Optional[List[int]] = None
  m_BitsPerSample: Optional[int] = None
  m_Channels: Optional[int] = None
  m_CompressionFormat: Optional[int] = None
  m_Format: Optional[int] = None
  m_Frequency: Optional[int] = None
  m_IsTrackerFormat: Optional[bool] = None
  m_Legacy3D: Optional[bool] = None
  m_Length: Optional[float] = None
  m_LoadInBackground: Optional[bool] = None
  m_LoadType: Optional[int] = None
  m_PreloadAudioData: Optional[bool] = None
  m_Resource: Optional[StreamedResource] = None
  m_Stream: Optional[int] = None
  m_SubsoundIndex: Optional[int] = None
  m_Type: Optional[int] = None
  m_UseHardware: Optional[bool] = None

  @property
  def extension(self) -> str: ...
  
  @property
  def samples(self) -> Dict[str, bytes]: ...
  
  ```
  extension = property(_AudioClip_extension)
  samples = property(_AudioClip_samples)

Nattsu39:
I think I need to click Re-request-review.
😶

PR #290, opened 2025-01-10
MeshHandler - Implement reading `m_StreamData`

mos9527:
Seems like it's possible to read external data in UnityPy for a while now (https://github.com/K0lb3/UnityPy/blob/master/UnityPy/export/AudioClipConverter.py#L97) so I've also implemented it here.
Tested on some bundles that do contain meshes using `m_StreamData` and it does seem to work well. I can attach the said bundle files here if needed.
This also addresses https://github.com/mos9527/sssekai_blender_io/issues/15


K0lb3 (approved):

K0lb3:
I will do a version bump later today or tomorrow.
Before the bump I have to fix the test workflow,.which currently fails as the set python version (3.7) was removed.

PR #292, opened 2025-01-26
Fix concurrent issue of ASTC context

isHarryh:
## Summary

This PR mainly fixes the concurrent issue of ASTC context in `Texture2DConverter`. By the way, it also fixes some type errors and type hints in `Texture2DConverter`.

### How do I discovered this issue:

After upgrading UnityPy (from 1.18 to 1.20) in my project [ArkUnpacker](https://github.com/isHarryh/Ark-Unpacker), I noticed that, when using multithreading to extract **images**, the application has an unpredictable chance to crash **without** any stacktrace (just like you shutdown the application in the Task Manager forcibly).

With the extensive debugging, I finally determined that it is the **asynchronous access to the ASTC context** that causes the problem.

This issue was caused by this commit d38b5f660d1c1f048413cb5dfa0464400f1038be which introduced `astc_encoder` for texture decoding.

### How to fix it:

Simply adding a `threading.Lock` to the dict `ASTC_CONTEXTS`.

## 2. Type Errors

There are two potential type errors. See my commit for detailed changes. Here comes the explanation:

1. In this case, `texture_format` may be an integer.

```python
        raise NotImplementedError(
-            f"Not implemented texture format: {texture_format.name}"
+            f"Not implemented texture format: {texture_format}"
        )
```

2. In this case, bytes cannot be divided by integer.

```python
    rgb_data = b"".join(
        stream.read(padding_size * 2) + padding
-        for _ in range(image_data / (2 * padding_size))
+        for _ in range(len(image_data) // (2 * padding_size))
    )
```

### DIFF
--- a/UnityPy/export/Texture2DConverter.py
+++ b/UnityPy/export/Texture2DConverter.py
@@ -1,8 +1,9 @@
-from __future__ import annotations
+﻿from __future__ import annotations
 
 import struct
 from copy import copy
 from io import BytesIO
+from threading import Lock
 from typing import TYPE_CHECKING, Dict, Tuple, Union
 
 import astc_encoder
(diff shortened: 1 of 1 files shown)


UnityPy/export/Texture2DConverter.py
```diff
@@ -74,22 +75,20 @@ def image_to_texture2d(
...
+            raw_img = astc_encoder.ASTCImage(
+                astc_encoder.ASTCType.U8, img.width, img.height, 1, raw_img
+            )
+            if img.mode == "RGB":
+                tex_format = getattr(TF, f"ASTC_RGB_{block_size[0]}x{block_size[1]}")
+            else:
+                tex_format = getattr(TF, f"ASTC_RGBA_{block_size[0]}x{block_size[1]}")
+
+            swizzle = astc_encoder.ASTCSwizzle.from_str("RGBA")
+            enc_img = context.compress(raw_img, swizzle)
```
K0lb3, line 90: The lock is only necessary for this line.

UnityPy/export/Texture2DConverter.py
```diff
@@ -166,7 +165,7 @@ def parse_image_data(
     texture_format: Union[int, TextureFormat],
     version: tuple,
     platform: int,
-    platform_blob: bytes = None,
+    platform_blob: Union[bytes, None] = None,
```
K0lb3, line 168: Please change the ``Union[foo, None]`` to ``Optional[foo]``, as that's what I've been using everywhere in UnityPy.

I think that Optional makes the implication more clear than a Union with None.

UnityPy/export/Texture2DConverter.py
```diff
@@ -177,7 +176,7 @@ def parse_image_data(
 
     if len(selection) == 0:
         raise NotImplementedError(
-            f"Not implemented texture format: {texture_format.name}"
+            f"Not implemented texture format: {texture_format}"
```
K0lb3, line 179: Good catch, I forgot that texture_format could still be an int.
As the int number isn't that telling imo, it would be better to cast the format to the enum and print it's name if possible, and only show the number if that is not possible.

UnityPy/export/Texture2DConverter.py
```diff
@@ -356,7 +363,7 @@ def rg(
     padding = bytes(padding_size)
     rgb_data = b"".join(
         stream.read(padding_size * 2) + padding
-        for _ in range(image_data / (2 * padding_size))
+        for _ in range(len(image_data) // (2 * padding_size))
```
K0lb3, line 366: 😓 ooph, thanks for catching that one, looks like I accidentally deleted my commit with the corrected version.

K0lb3 (changes requested):
Thanks for the PR and the fixes.
I just want to see two little changes overall, reducing the lock scope to the specific context usage and replacing Union[foo, None] with Optional[foo].

UnityPy/export/Texture2DConverter.py
isHarryh: If the int can be cast to the enum, it is very likely to be a known format that has been implemented. In other words, the unimplemented texture format is very likely to be not castable. So showing the int value is enought.

K0lb3: Not all of them are implemented atm, but so far I haven't seen any of those in the wild either.
So I guess it's indeed fine to keep it as is as int.

K0lb3 (approved):

K0lb3:
Thanks a lot 👍

PR #293, opened 2025-01-30
Fix some type hints (EndianBinaryReader, EndianBinaryWriter)

isHarryh:
## Summary

This PR mainly fixes some type hints in `EndianBinaryReader` and `EndianBinaryWriter`.

By the way, some minor code issues are fixed, including:

1. Legacy code lines (e.g. legacy `pass` statement) are removed.
2. Some imports and isinstance-checks are optimized.
3. Some potential errors (e.g. int conversion) are fixed.

For more details, please refer to the commits.

### DIFF
--- a/UnityPy/streams/EndianBinaryReader.py
+++ b/UnityPy/streams/EndianBinaryReader.py
@@ -1,8 +1,10 @@
 import sys
-from struct import Struct, unpack
 import re
-from typing import List, Union
-from io import BytesIO, BufferedIOBase, IOBase, BufferedReader
+from struct import Struct, unpack
+from io import IOBase, BufferedReader
+
+import builtins
+from typing import Callable, List, Optional, Tuple, Union
 
 reNot0 = re.compile(b"(.*?)\x00", re.S)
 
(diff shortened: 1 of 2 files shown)


K0lb3 (dismissed):
Thanks a lot 👍

K0lb3:
What is the point of ``builtins.bytes`` though?
Isn't it the same as the normal bytes?

isHarryh:
> What is the point of `builtins.bytes` though? Isn't it the same as the normal bytes?

Here is a minimal example shows why `bytes` without `builtins` doesn't work:

```python
class EndianBinaryReader:
    endian: str
    ...

    def __new__(cls, ...):
        ...

    def __init__(self, ...):
        ...

    @property
    def bytes(self):
        # implemented by Streamable and Memoryview versions
        return b""

    ...

    def read_bytes(self, num) -> builtins.bytes:
        return self.read(num)

    ...
```

In the code above, `bytes` was first declared as a property name. This will overwrite the original `bytes` (the type name).

So, if the type hint still uses `bytes`, the linter will recognize it as the property name `bytes`, which will cause some type hint error.

image

To avoid this problem, it's recommended to use `builtins.bytes` in the type hint.

isHarryh:
There are some changes (in `EndianBinaryReader`). I have force-pushed these changes.

K0lb3:
> In the code above, bytes was first declared as a property name. This will overwrite the original bytes (the type name).
I see, makes sense. So far it worked fine for me without issues, but I'm not against using this fix if it's causing issues for you and potentially others as well.

> Why you use self._endian = ""? Is this intentional? Why not self._endian = endian?
``@endian`` is a property that sets "optimized" functions. As it checks for a difference in endian at that point, the initial ``_endian`` value has to be set to an unaccepted value, so that the ``.endian`` property call in ``__init__`` can detect an endian change, thereby setting the correct endianed reader functions.

It's certainly not a great solution and could be done better.
Assigning endianed functions yields a sizeable performance increase though, which is why I'm sticking with it so far.
For the next minor update I already made a small rewrite for this, which assigns functions instead of overwriting the ``__class__``, as the later isn't allowed in the latest python versions anymore.

K0lb3 (approved):

PR #294, opened 2025-01-31
Refactor the math module

isHarryh:
## Summary

This PR has undergone comprehensive optimizations and fixes for the **math** module.

## Breaking Changes

1. Complete implementation for `Vector2` and `Vector4`. (According to `Vector3`.)
2. Remove the file `Half.py` since the only function `ToHalf` in this file is not used any more. (`ToHalf` was used in the `Texture2DConvertor` in the earlier version of UnityPy, but this usage has been removed in later versions.)

## Common Changes

1. Type hints.
2. Use `@dataclass` decorator for all classes. (Before this PR, only `Vector3` had this decorator.)
3. Use `__truediv__` (Python 3) to replace `__div__` (Python2).
4. Argument type validation.
5. Avoid duplicate `__eq__` method declarations.
6. Remove unnecessary comments at the end of the file.

## Individual Optimizations

1. `Vector3`: Use `Method = method` to redirect duplicated method.
2. `Vector3`: Use a better `__eq__` method (which is also [Unity's implementation](https://github.com/Unity-Technologies/UnityCsReference/blob/master/Runtime/Export/Math/Vector3.cs#L463-L471)).
3. `Quaternion`: Remove the redundant `self._data` field.
4. `Color`: Use a better `__eq__` method.

## Individual Fixes

1. `Vector3`: Incorrect [if-statement](https://github.com/K0lb3/UnityPy/blob/master/UnityPy/math/Vector3.py#L15) in `__init__` method.
2. `Vector3`: Incorrect [else-body](https://github.com/K0lb3/UnityPy/blob/master/UnityPy/math/Vector3.py#L51-L53) in `normalize` method.
3. `Vector3`: Incorrect `LengthSquare` method.
4. `Quaternion`: The `w` value of the [identity quaternion](https://github.com/Unity-Technologies/UnityCsReference/blob/efc9c4d56674f0ebec477747863562224da2f541/Runtime/Export/Math/Quaternion.cs#L73) (default quaternion) is `1.0`, not `0.0`.
6. `Matrix4x4`: Corrupted `__eq__` method.


UnityPy/math/Vector4.py
```diff
@@ -1,155 +1,112 @@
...
+        if len(args) == 1:
+            args = args[0]
+
+        if isinstance(args, Sequence):
+            if len(args) == 4:  # args=(x, y, z, w)
+                self.X, self.Y, self.Z, self.W = args
+                return
+            if len(args) == 2:  # args=(Vector3, w)
+                self.X, self.Y, self.Z = args[0]
+                self.W = args[1]
```
K0lb3, line 28: A return is missing here.
I guess that a simple 
if, elif, ..., ellif raise TypeError would be cleaner

K0lb3 (changes requested):
Just one change

K0lb3:
Thanks for the PR and fixes.

I'm also rewriting and pruning these classes for the next minor patch,
so you might have ended up wasting some time, sorry for taking a bit too long with that patch.

For the version 2 update I plan to include numpy....and for that the math classes are going to get another rewrite,
subtyping them from the np.ndarray class, making them directly compatible with numpy functions.

But that's still some time off,
as I will have to focus on other things in the next months.
So thanks a lot for these changes and fixes.

isHarryh:
No problem at all! The simple work didn’t take long, happy to help. Good luck with the UnityPy 2 updates!

PR #295, opened 2025-02-01
tools/extractor.py: only access .path_id and .type.name on object reader

michaelskyba:
Currently there is an `obj = obj.read()` overwrite that changes the object reader to a direct object, but some other lines still expect the original object reader


K0lb3 (approved):
Thanks a lot 👍

PR #297, opened 2025-02-06
Improve helpers module

isHarryh:
## Summary

This PR does some minor improvements on the helpers module. These improvements will not affect the existing features.

## Common Changes

1. Type hints.

## Individual Changes

1. `ArchiveStorageDecryptor`: Uses an inline approach to replace the function `to_uint4_array`, which can improve the performance and simplify the code.
2. `ResourceReader`: Refactors the `get_resource_data` function, because this function used an implicit argument overloading trick (mixed the 4-args invocation and the 3-args invocation, but only 4-args invocation is used externally), which is bad for maintaining.
3. `UnityVersion`: Fixes the error in the method `fromString`.

### DIFF
--- a/UnityPy/helpers/MeshHelper.py
+++ b/UnityPy/helpers/MeshHelper.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
+
 import math
 import struct
-from typing import Optional, List, Tuple, Union, TypeVar
+from typing import Optional, List, Sequence, Tuple, Union, TypeVar
 
-from ..enums.MeshTopology import MeshTopology
 from ..classes.generated import (
     ChannelInfo,
     StreamInfo,
@@ -13,8 +13,7 @@ from ..classes.generated import (
     Vector3f,
     Vector4f,
 )
-from .PackedBitVector import unpack_floats, unpack_ints
-
+from ..enums.MeshTopology import MeshTopology
 from ..enums.VertexFormat import (
     VertexChannelFormat,
     VertexFormat2017,
@@ -23,6 +22,7 @@ from ..enums.VertexFormat import (
     VERTEX_FORMAT_2017_STRUCT_TYPE_MAP,
     VERTEX_FORMAT_STRUCT_TYPE_MAP,
 )
+from .PackedBitVector import unpack_floats, unpack_ints
 from .ResourceReader import get_resource_data
 
 try:
@@ -37,11 +37,11 @@ Tuple4f = Tuple[float, float, float, float]
 T = TypeVar("T")
 
 
-def flat_list_to_tuples(data: List[T], item_size: int) -> List[tuple[T]]:
+def flat_list_to_tuples(data: Sequence[T], item_size: int) -> List[tuple[T, ...]]:
     return [tuple(data[i : i + item_size]) for i in range(0, len(data), item_size)]
 
 
-def vector_list_to_tuples(data: List[Vector2f, Vector3f, Vector4f]) -> List[tuple]:
+def vector_list_to_tuples(data: List[Union[Vector2f, Vector3f, Vector4f]]) -> List[tuple]:
     if isinstance(data[0], Vector2f):
         return [(v.x, v.y) for v in data]
     elif isinstance(data[0], Vector3f):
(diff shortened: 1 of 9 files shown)


UnityPy/helpers/MeshHelper.py
```diff
@@ -97,12 +97,12 @@ def __init__(
         src: Union[Mesh, SpriteRenderData],
         version: Optional[Tuple[int, int, int, int]] = None,
         endianess: str = "<",
-    ) -> None:
+    ):
         self.src = src
         self.endianess = endianess
         if version is not None:
             self.version = version
-        elif getattr(src, "object_reader", None):
```
K0lb3, line 105: I used ``getattr`` here, as ``object_reader`` can be ``None``.
``hasattr`` would return ``True`` if it's ``None``.

UnityPy/helpers/Tpk.py
```diff
@@ -185,7 +185,7 @@ def GetDataBlob(self) -> TpkDataBlob:
         elif self.CompressionType == TpkCompressionType.Brotli:
             import brotli
 
-            decompressed: bytes = brotli.decompress(self.CompressedBytes)
+            decompressed = brotli.decompress(self.CompressedBytes)
```
K0lb3, line 188: Google still hasn't merged my PR that would add type hints to brotli,
so typehinters should expect an "any" or "unknown" value for decompressed.
Either ``:bytes`` or ``= cast(bytes, ...)`` has to be used.

UnityPy/helpers/Tpk.py
```diff
@@ -4,11 +4,11 @@
 from importlib.resources import open_binary
 from io import BytesIO
 from struct import Struct
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 from .TypeTreeHelper import TypeTreeNode
 
-TPKTYPETREE: TpkTypeTreeBlob = None
+TPKTYPETREE: Optional[TpkTypeTreeBlob] = None
```
K0lb3, line 11: TPKTYPETREE is always a TpkTypeTreeBlob due to the init call at the end of the file.
So the Optional would add more trouble, as the other functions here would then need  unnecessary None checks.

UnityPy/helpers/ResourceReader.py
```diff
@@ -1,51 +1,33 @@
 import ntpath
-from ..streams import EndianBinaryReader
-from ..files import File
+from ..files.SerializedFile import SerializedFile, EndianBinaryReader
```
K0lb3, line 2: I'm a bit touchy with SerializedFile due to recursive import issues.
As long as it's only used for type-hinting it should be done with ``if TYPE_CHECKING``.
EndianBinaryReader should also be imported from streams and not indirectly in this way.

UnityPy/helpers/MeshHelper.py
isHarryh: 😓 ooph, sorry for introducing a bug.

UnityPy/helpers/Tpk.py
isHarryh: I see.
I guess it would be better if we use something like `TPKTYPETREE: TpkTypeTreeBlob = init()`, which is also an execute-at-import behavior but result in more clear intention and better type checking experience which avoid separating the `init()` call to the end of the file.

isHarryh:
According to your 4 reviews. I have changed these things and force-pushed these changes.

UnityPy/helpers/Tpk.py
```diff
@@ -4,22 +4,23 @@
...
-
-    global TPKTYPETREE
     with BytesIO(data) as stream:
-        TPKTYPETREE = TpkFile(stream).GetDataBlob()
+        blob = TpkFile(stream).GetDataBlob()
+        assert isinstance(blob, TpkTypeTreeBlob)
+        return blob
+
+
+TPKTYPETREE: TpkTypeTreeBlob = init_tpk_typetree()
```
K0lb3, line 21: This doesn't work as TpkFile isn't defined yet on function call.
It simply errors.

```
Exception has occurred: NameError
name 'TpkFile' is not defined
```

UnityPy/helpers/Tpk.py
K0lb3: Sadly that doesn't work, as the parser classes aren't defined yet at that point.
It simply errors.

```
Exception has occurred: NameError
name 'TpkFile' is not defined
```

isHarryh:
Okay. 2 fixes were applyed since the last push:

1. Revert the changes in `Tpk.py` which caused `NameError`.
2. Use forward reference `"SerializedFile"` in `ResourceReader`.

K0lb3 (approved):
Looking good now, thanks

PR #299, opened 2025-02-11
TypeTreeHelper - Add missing `size` argument in `read_value_array`

mos9527:
Title. The Python TypeTree reader implementation omitted `size` in this particular code path.


K0lb3 (approved):
Thanks for the fix and type fix^^

PR #303, opened 2025-02-22
Improve config and files module

isHarryh:
## Summary

This PR makes code quality improvements to the files module, such as type hints and typo corrections.

### 1. fix: better type hint in EndianBinaryReader/Writer

In `EndianBinaryReader`, use `-> Tuple[xxx, ...]` instead of `-> Tuple[xxx]`, where the latter one is a common mistake.

In `EndianBinaryWriter`, use `value: Sequence` instead of `value: list` because some invocations pass a tuple argument instead of list.

### 2. refactor(files): use assertion for version-specific assignment

Use `assert xxx is not None` before value assignment.

This is also the practice adopted by `MeshHelper` and other classes, which can correctly satisfy the NoneType checking.

### 3. chore(files): fix typo and remove legacy code

In `File` class:
- `class File(object)` can be simplified to `class File` (Python 3).
- `self.environment = self.environment = ...` causes a duplicated self-assignment.

In `ContainerHelper`, method `__or__` cannot be used anymore since 5ac0a62c59102ee0c53ad265d7d589d6ad6e7ccd changed the constructor of this class.

### 4. fix(files): overall type hint corrections

Correct the type hints in the files module.

Refactors `BundleFile.get_version_tuple`:
- Uses `(next(...), next(...), next(...))` instead of `tuple(map(...))` to respect the returned type `-> Tuple[int, int, int]`.
- Prevents `AttributeError` by validating the regex match before accessing `groups`.

> **Updates with 2 commits:**  
> 
> ### 1. fix(files): use generic type in EndianBinaryWriter
>
> Applies generic type to `Sequence` and `Callable` in `EndianBinaryWriter`.
> ### 2. feat(config): unset default fallback version
>
> Sets the `FALLBACK_UNITY_VERSION` to `None` (unset).
>
> Raises `UnityVersionFallbackError` when the fallback version is unset or invalid.
>
> Refactors the warning mechanism using the `warnings` module.
>
> Optimizes the documents string of the constants in the config.

### DIFF
--- a/UnityPy/files/BundleFile.py
+++ b/UnityPy/files/BundleFile.py
@@ -5,6 +5,7 @@ from typing import Optional, Tuple, Union
 
 from . import File
 from ..enums import ArchiveFlags, ArchiveFlagsOld, CompressionFlags
+from ..exceptions import UnityVersionFallbackError
 from ..helpers import ArchiveStorageManager, CompressionHelper
 from ..streams import EndianBinaryReader, EndianBinaryWriter
 
@@ -531,6 +532,6 @@ class BundleFile(File.File):
         if not match:
             match = reVersion.match(config.get_fallback_version())
             if not match or len(match.groups()) < 3:
-                raise ValueError("Illegal fallback version format")
+                raise UnityVersionFallbackError("Illegal fallback version format")
         map_ = map(int, match.groups())
         return (next(map_), next(map_), next(map_))
--- a/UnityPy/files/BundleFile.py
+++ b/UnityPy/files/BundleFile.py
@@ -201,8 +201,8 @@ class BundleFile(File.File):
                 original - uses the original flags
         """
         # file_header
-        #     signature    (string_to_null)
-        #     format        (int)
+        #     signature         (string_to_null)
+        #     format            (int)
         #     version_player    (string_to_null)
         #     version_engine    (string_to_null)
         writer = EndianBinaryWriter()
@@ -252,11 +252,11 @@ class BundleFile(File.File):
         # data_flag
 
         # header:
-        #     bundle_size        (long)
-        #     compressed_size    (int)
-        #     uncompressed_size    (int)
-        #     flag                (int)
-        #     ?padding?            (bool)
+        #     bundle_size       (long)
+        #     compressed_size   (int)
+        #     uncompressed_size (int)
+        #     flag              (int)
+        #     ?padding?         (bool)
         #   This will be written at the end,
         #   because the size can only be calculated after the data compression,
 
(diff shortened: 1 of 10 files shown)


UnityPy/files/BundleFile.py
```diff
@@ -523,6 +523,14 @@ def decompress_data(
...
         """Returns the version as a tuple."""
         version = self.version_engine
-        if not version or version == "0.0.0":
-            version = config.get_fallback_version()
-        return tuple(map(int, reVersion.match(version).groups()))
+        match = None
+        if version and version != "0.0.0":
+            match = reVersion.match(version)
+            if not match or len(match.groups()) < 3:
+                match = None
```
K0lb3, line 530: In this case it might actually be better to throw an error if the fallback version isn't set.

K0lb3: ....which requires the fallback version to not be set by default...which would be better than using an old version as of now anyway.

UnityPy/streams/EndianBinaryWriter.py
```diff
@@ -154,29 +154,29 @@ def write_byte_array(self, value: builtins.bytes):
         self.write_int(len(value))
         self.write(value)
 
-    def write_boolean_array(self, value: list):
+    def write_boolean_array(self, value: Sequence):
```
K0lb3, line 157: The sequence type could be specified here.

UnityPy/streams/EndianBinaryWriter.py
```diff
@@ -144,7 +144,7 @@ def write_matrix(self, value: Matrix4x4):
         for val in value.M:
             self.write_float(val)
 
-    def write_array(self, command: Callable, value: list, write_length: bool = True):
+    def write_array(self, command: Callable, value: Sequence, write_length: bool = True):
```
K0lb3, line 147: callable and sequence could use a Generic to make it more explicit,
but that might not be worth the trouble and potential conflicts.

UnityPy/files/BundleFile.py
K0lb3: I can also just change that with a follow up commit, so it can stay this way.

UnityPy/streams/EndianBinaryWriter.py
isHarryh: Do you mean "the sequence type should only be specified here" ?

From my view, all the `write_xxx_array` method should have their `value: list` to be `value: Sequence`.

K0lb3: I meant something like
Sequence[bool] for write_boolean_array,
Sequence[int] for all the int array one,
and so on

K0lb3: I totally agree to changing it from list to Sequence

isHarryh: Okay, sorry for my bad English comprehension. XD

Not a native speaker.

UnityPy/streams/EndianBinaryWriter.py
isHarryh: Okay. Your suggestion is good. It worth the trouble anyway.

Issue #308, opened 2025-04-07
Support LZHAM / custom compression


isHarryh:
### Description

The game [Arknights](https://ak.hypergryph.com) \(CN server\) has upgraded to v2.5.04 today. In this version, they upgraded the Unity engine from `2017.4.39f1` to `2021.3.39f1`. More importantly, they adopted a **LZHAM** compression that both UnityPy and AssetStudio not implemented yet.

### Stacktrace

When invoke `UnityPy.load()`:

```txt
LZHAM decompression not implemented
  File "xxx.py", line xxx, in __init__
    self._env = UnityPy.load(path)
```

### Reproduction

Using UnityPy v1.21.3 (latest).

Here are some sample files: [ak2504samples.zip](https://github.com/user-attachments/files/19625872/ak2504samples.zip)

K0lb3:
LZHAM is basically dead and not included in Unity anymore, including the old version where it was originally supposed to be used (around Unity 5).
When extractors detect LZHAM it's a sign that a custom compression gets used, which is most of the time just an encryption on top of lz4.

isHarryh:
I am wondering that can we add a function like `set_fallback_decompression_method` to `helpers/CompressionHelper.py` ?

For example:

```py
FALLBACK_DECOMPRESS = None

def set_fallback_decompression_method(method: Callable[[bytes, int], bytes]):
    FALLBACK_DECOMPRESS = method
```

then in `files/BundleFile.py`:

```diff
-        elif comp_flag == CompressionFlags.LZHAM:  # LZHAM
-            raise NotImplementedError("LZHAM decompression not implemented")
+        else:
+            if CompressionHelper.FALLBACK_DECOMPRESS:
+                return CompressionHelper.FALLBACK_DECOMPRESS(compressed_data, uncompressed_size)
+            raise NotImplementedError(f"{comp_flag} decompression not implemented")
```

This will enable users to easily custom the block decompression, since many games use the custom decompression (like Bu Liang Ren 3 mentioned in #301 and Arknights mentioned in this issue, etc).

What is your opinion?

K0lb3:
You can already hack around the issue by overwriting the decompress_data function of the BundleFile class.
e.g.
```py
from UnityPy.files.BundleFile import BundleFile

BundleFile.decompress_data = your_custom_function
```
I concede that it's not a great solution, but it works.


Your idea isn't bad and is in line with how I handle the Unity fallback version,
[...]

PR #309, opened 2025-04-08
use maps for compression and decompression functionality

K0lb3:
This PR adds a mapping layer for the compression and decompression functions used in BundleFile.
This layer makes it easier to use custom decompression methods for games that require it.

It solves #308

### DIFF
--- a/UnityPy/files/BundleFile.py
+++ b/UnityPy/files/BundleFile.py
@@ -1,7 +1,7 @@
 # TODO: implement encryption for saving files
 import re
 from collections import namedtuple
-from typing import Optional, Tuple, Union
+from typing import Optional, Tuple, Union, cast
 
 from .. import config
 from ..enums import ArchiveFlags, ArchiveFlagsOld, CompressionFlags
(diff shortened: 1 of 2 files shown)


UnityPy/helpers/CompressionHelper.py
```diff
@@ -184,34 +199,92 @@ def chunk_based_compress(data: bytes, block_info_flag: int) -> Tuple[bytes, list
...
+        "Custom compression or unimplemented LZHAM (removed by Unity) encountered!"
+    )
+
+
+DECOMPRESSION_MAP: Dict[
+    Union[int, CompressionFlags], Callable[[ByteString, int], ByteString]
+] = {
+    CompressionFlags.NONE: lambda cd, _ucs: cd,
+    CompressionFlags.LZMA: lambda cd, _ucs: decompress_lzma(cd),
+    CompressionFlags.LZ4: lambda cd, ucs: decompress_lz4(cd, ucs),
```
isHarryh, line 254: Use function name as value directly, like `CompressioFlags.xxx: decompress_xxx`, may be better.

K0lb3: This only works for decompress_lz4, besides if I would change the signature of decompress_lzma to include an unused value.
But yeah, it should be done that way for lz4.

K0lb3: The same also applies to the compression functions below.

K0lb3: Done, and thanks for mentioning it.

UnityPy/files/BundleFile.py
```diff
@@ -525,16 +525,20 @@ def decompress_data(
...
+        if comp_flag in CompressionHelper.DECOMPRESSION_MAP:
+            return cast(
+                bytes,
+                CompressionHelper.DECOMPRESSION_MAP[comp_flag](
+                    compressed_data, uncompressed_size
+                ),
+            )
         else:
-            return compressed_data
+            raise ValueError(
```
isHarryh, line 539: Raise `NotImplementedError` here may be more informative.

isHarryh (approved):
It worked well and is ready to merge.

Thanks a lot. 👍

PR #310, opened 2025-04-11
Documentation correction and update

isHarryh:
### Summary

This PR:

- Fixes some spelling typo.
- Removes the 404 links in the "Important Object Types" (they have been moved to `legacy_patch`).
- Addes a "Configurations" section where user can have a full understanding of customization.
- Moves "UnityPy Encryption" and "C-Typetree" related content to the "Configuration" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ if UnityPy.__version__ != '1.9.6':
 2. [Example](#example)
 3. [Important Classes](#important-classes)
 4. [Important Object Types](#important-object-types)
-5. [Custom Filesystem](#custom-filesystem)
+5. [Configurations](#configurations)
 6. [Credits](#credits)
 
 ## Installation
@@ -73,15 +73,10 @@ In case a new(ish) Python version is used, it can happen that the C-dependencies
 In such cases the user either has to report this as issue or follow the steps of [this issue](https://github.com/K0lb3/UnityPy/issues/223) to compile it oneself.
 Another option for the user is downgrading Python to the latest version supported by UnityPy. For this see the Python version badge at the top of the README.
 
-### Crash without warning/error
+#### Crash without warning/error
 
 The C-implementation of the typetree reader can directly crash Python.
-In case this happens, the usage of the C-typetree reader can be disabled by adding these two lines to your main file.
-
-```python
-from UnityPy.helpers import TypeTreeHelper
-TypeTreeHelper.read_typetree_boost = False
-```
+In case this happens, the usage of the C-typetree reader can be disabled. Read [this section](#disable-typetree-c-implementation) for more details.
 
 ## Example
 
@@ -135,13 +130,8 @@ def unpack_all_assets(source_folder: str, destination_folder: str):
 You probably have to read [Important Classes](#important-classes)
 and [Important Object Types](#important-object-types) to understand how it works.
 
-People with slightly advanced Python skills should look at [UnityPy/tools/extractor.py](UnityPy/tools/extractor.py) for a more advanced example.
+Users with slightly advanced Python skills should look at [UnityPy/tools/extractor.py](UnityPy/tools/extractor.py) for a more advanced example.
 It can also be used as a general template or as an importable tool.
-
-### Setting the decryption key for Unity CN's AssetBundle encryption
-
-The Chinese version of Unity has its own builtin option to encrypt AssetBundles/BundleFiles. As it's a feature of Unity itself, and not a game specific protection, it is included in UnityPy as well.
-To enable encryption simply use `UnityPy.set_assetbundle_decrypt_key(key)`, with key being the value that the game that loads the bundles passes to `AssetBundle.SetAssetBundleDecryptKey`.
 
 ## Important Classes
 
@@ -350,7 +340,7 @@ The samples are converted into the .wav format.
 The sample data is a .wav file in bytes.
 
 ```python
-clip : AudioClip
+clip: AudioClip
 for name, data in clip.samples.items():
     with open(name, "wb") as f:
         f.write(data)
@@ -362,7 +352,7 @@ for name, data in clip.samples.items():
 
 ```python
 if obj.type.name == "Font":
-    font : Font = obj.read()
+    font: Font = obj.read()
     if font.m_FontData:
         extension = ".ttf"
         if font.m_FontData[0:4] == b"OTTO":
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,11 +25,11 @@ Via the typetree structure all objects types can be edited.
 If you need advice or if you want to talk about (game) data-mining,
 feel free to join the [UnityPy Discord](https://discord.gg/C6txv7M).
 
-If you're using UnityPy a commercial project,
-a donation to a charitable cause or a sponsorship of this project is expected.
+If you're using UnityPy for a commercial project,
+a donation to a charitable cause or a sponsorship of this project will be appreciated.
```
3ncy, line 29: This significantly changes the meaning

isHarryh: Reverted.

README.md
```diff
@@ -183,26 +174,26 @@ with open(dst, "wb") as f:
...
 
-Objects contain the _actual_ files, e.g., textures, text files, meshes, settings, ...
+Objects \([ObjectReader class](UnityPy/files/ObjectReader.py)\) contain the _actual_ files, e.g., textures, text files, meshes, settings, ...
 
 To acquire the actual data of an object it has to be read first. This happens via the `.read()` function. This isn't done automatically to save time because only a small part of the objects are of interest. Serialized objects can be set with raw data using `.set_raw_data(data)` or modified with `.save()` function, if supported.
 
 ## Important Object Types
 
-All object types can be found in [UnityPy/classes](UnityPy/classes/).
+Now UnityPy uses auto generated classes while all object types can also be found in [UnityPy/classes/legacy_patch/](UnityPy/classes/legacy_patch/).
```
3ncy, line 194: I feel like this is a bit misleading, because the `legacy_patch/` folder contains only few select types with specific overridden behaviour, for example the `get_image_data()` method on Texture2D. The declaration of all types (including those in `legacy_patch/` is still within the `classes/` folder, in `generated.py`.

Or is this edit in anticipation of future changes where the individual class files will be deprecated?

README.md
```diff
@@ -183,26 +174,26 @@ with open(dst, "wb") as f:
...
 
-### [Texture2D](UnityPy/classes/Texture2D.py)
+### Texture2D
```
3ncy, line 196: Wouldn't it be better to correct the link, to make it point to the `classes/legacy_patch/` folder, instead of removing it?

README.md
isHarryh: You're right, this statement may make people mistakenly think that these contents will be deprecated. Thanks for mentioning.

README.md
3ncy: Yep, that sounds like a good idea to point them to the `.pyi` files in `legacy_patch`.
The only objects of these, that don't have their own hand-written implementation, are `TextAsset`, `MonoBehaviour` and `Font`. I guess they can be left without a link then, unless you think it's a good idea to link to `generated.py` for these

K0lb3:
I will take a look at this PR tomorrow.
I had some important stuff to handle IRL, and afterwards some tasks that accumulated due to it.

From a glance the changes look fine, but as @3ncy already mentioned in his comments, some rephrasing change the meanings of the text, which has to be discussed.

README.md
K0lb3: Long term I plan to move all legacy patches into separate, optional plugins, that act like a wrapper around the specific classes instead of extending them.

So just linking to ``generated.py`` is fine.


The existence of ``legacy_patch`` and that it extends some classes would make sense tho.

README.md
```diff
@@ -423,32 +416,76 @@ for obj in env.objects:
...
+### Custom Block (De)compression
+
+Some game assets have non-standard compression/decompression algorithm applied on the block data. If you wants to customize the compression/decompression function, you can modify the corresponding function mapping. e.g.
+
+```python
+from UnityPy.enums.BundleFile import CompressionFlags
+flag = CompressionFlags.LZHAM
+
+from UnityPy.helpers import CompressionHelper
+CompressionHelper.COMPRESSION_MAP[flag] = some_function
```
K0lb3, line 460: Instead of some or another it's probably better to simply name them

custom_compression_function and custom_decompression function,
with a stub above to show their expected functions signature.

README.md
3ncy: With the current state of the documentation, I think it is important to mention `legacy_patch`, because otherwise people might not even discover it, or know that is being used (due to its name).

README.md
```diff
@@ -9,8 +9,9 @@
 A Unity asset extractor for Python based on [AssetStudio](https://github.com/Perfare/AssetStudio).
 
 Next to extraction, UnityPy also supports editing Unity assets.
-Via the typetree structure all objects types can be edited.
-```py
+Via the typetree structure all object types can be edited.
```
K0lb3, line 12: It might make sense to say ``can be edited in their native forms`` and then proceeding to say that some have handled export functions to make the export easier.

K0lb3:
Sorry for the late response, I kinda forgot about the PR, and well, then Easter came with lots of hiking.

isHarryh:
> Sorry for the late response, I kinda forgot about the PR, and well, then Easter came with lots of hiking.

That's okay. Btw, congratulations to UnityPy for receiving 1k stars!

I have updated the commits according to the suggestions from you and @3ncy . You can view the [diffs](https://github.com/K0lb3/UnityPy/compare/738ea76c845fffa95a50a581e20e2e0a8e8d6d3c..cd112a3e6cabeb61f48b201a2d478a5ccfe4e105).

README.md
```diff
@@ -246,14 +237,14 @@ for obj in env.objects:
...
-### [TextAsset](UnityPy/classes/TextAsset.py)
+### TextAsset
 
 TextAssets are usually normal text files.
 
 -   `.m_Name`
 -   `.m_Script` - str
 
-Some games save binary data as TextFile, so to convert the ``str`` back to bytes correctly ``m_Script.encode("utf-8", "surrogateescape")`` has to be used.
+Some games save binary data as text file, so to convert the ``str`` back to bytes correctly ``m_Script.encode("utf-8", "surrogateescape")`` has to be used.
```
K0lb3, line 247: Only noticed this issue now.
I meant TextAsset in the original text, not text file.

The framing could also be improved a little bit overall.

Some games save binary data as TextFile, so to convert the ``str`` back to bytes correctly use ``m_Script.encode("utf-8", "surrogateescape")``.

K0lb3 (changes requested):
Again, thanks a lot.
Just one thing that I only noticed now has to be changed.

README.md
isHarryh: I just thought that `TextFile` was a typo of `text file`, since there are no TextFile class.

You are right, change it to `TextAsset` is better.

K0lb3 (approved):

K0lb3:
Thanks a lot again 🙏

PR #318, opened 2025-05-26
Fix Switch swizzle texture conversion and optimize Texture2DConverter

### LINKED ISSUE #317: Export Texture2D wrong for game TokimekiMemorial1 switch
**Code**
```python
import os
import UnityPy

print(UnityPy.__version__)
env = UnityPy.load("070ebd75fa14728b6efa688dcaf25b3ad6b53f60122957e823ae43d0d863df47.assetbundle")
for obj in env.objects:
    if obj.type.name == "Texture2D":
        data = obj.read()
        path = os.path.join("", f"{data.m_Name}.png")
        try:
            data.image.save(path)
            print(obj.container, obj.path_id, data.m_Name)
        except Exception as e:
            pass
```

```
Output:
1.22.3
assets/app/assetbundledatas/global/font/df-tokubutogothic-w10001.ttf 2743269464722013593 Font Texture
assets/app/assetbundledatas/global/font/psfont.png 3956642238002007481 PSFont
```

**Error**
The exported texture psfont is as below
Image

**To Reproduce**
Here's the example asset bundle file. 
[070ebd75fa14728b6efa688dcaf25b3ad6b53f60122957e823ae43d0d863df47.assetbundle.gz](https://github.com/user-attachments/files/20427098/070ebd75fa14728b6efa688dcaf25b3ad6b53f60122957e823ae43d0d863df47.assetbundle.gz)

- following data:
  - Python version 3.8.15
  - UnityPy version 1.22.3

### LINKED ISSUE #319: Export Texture2D swizzle problem for TokimekiGirlSide3 switch
Similar as previous issue https://github.com/K0lb3/UnityPy/issues/317. 
I also find another game  texture2d not correct as in DXT5 format. 
Here's the sample and preview. 

[command_icon.assetbundle.gz](https://github.com/user-attachments/files/20455823/command_icon.assetbundle.gz)

Image

Image

@isHarryh

isHarryh:
### Summary

This PR mainly fixes the issue of **Switch** texture decoding, including:

- Fixes Alpha8 swizzle texture decoding and encoding.
- Fixes RGB24 swizzle texture decoding and encoding.
- Fixes DXT5 swizzle texture decoding and encoding.
- Fixes incorrect swizzle detection (platform blob).

This PR also does some improvements and small fixes to the Texture2DConverter:

- Removes the unnecessary `copy` procedure in `parse_image_data`, making it ~5% faster.
- Removes the unnecessary `paste` procedure in `pad_image`, making it ~20% faster.
- Fixes an error in `swap_bytes_for_xbox` that `bytes` object is not mutable.
- Changes `A` mode to `L` mode in `image_to_texture2d`, because `A` is not a valid PIL image mode.
- Refactors the `CONV_TABLE`.
  - Now `TextureFormat -> (converter function, (additional args, ...))`.
  - Typo of `EAC_R_SIGNED` has been fixed.
  - ETC series uses `"ETC2_RGB"`-like fmt instead of `[2, "RGB"]`-like.
- Wraps the functions in the TextureSwizzler.

### Related Issues

Fixes #317

Fixes #319

### DIFF
--- a/UnityPy/export/Texture2DConverter.py
+++ b/UnityPy/export/Texture2DConverter.py
@@ -3,7 +3,7 @@
 import struct
 from io import BytesIO
 from threading import Lock
-from typing import TYPE_CHECKING, Callable, Dict, Optional, Sequence, Tuple, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Sequence, Tuple, Union
 
 import astc_encoder
 import texture2ddecoder
@@ -327,12 +327,12 @@ def parse_image_data(
         else:
             image_data = texture2ddecoder.unpack_crunch(image_data)
 
-    conv_func = CONV_TABLE.get(texture_format)
-    if not conv_func:
+    if texture_format not in CONV_TABLE:
         raise NotImplementedError(
             f"Not implemented texture format: {texture_format.name}"
         )
-    img = conv_func(image_data, width, height)
+    conv_func, conv_args = CONV_TABLE[texture_format]
+    img = conv_func(image_data, width, height, *conv_args)
 
     if ori_width != width or ori_height != height:
         img = img.crop((0, 0, ori_width, ori_height))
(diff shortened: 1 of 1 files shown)


UnityPy/export/Texture2DConverter.py
```diff
@@ -450,22 +460,22 @@ def pvrtc(image_data: bytes, width: int, height: int, fmt: bool) -> Image.Image:
     return Image.frombytes("RGBA", (width, height), image_data, "raw", "BGRA")
 
 
-def etc(image_data: bytes, width: int, height: int, fmt: list) -> Image.Image:
-    if fmt[0] == 1:
+def etc(image_data: bytes, width: int, height: int, fmt: str) -> Image.Image:
```
isHarryh, line 463: Here we changed the `fmt` arg of `etc` from sth like `[2, 'RGB']` to `ETC2_RGB` thus maintaining consistency with `eac`.

UnityPy/export/Texture2DConverter.py
```diff
@@ -223,24 +222,27 @@ def image_to_texture2d(
...
+    if platform == BuildTarget.Switch and platform_blob:
+        gobs_per_block = TextureSwizzler.get_switch_gobs_per_block(platform_blob)
+
         s_tex_format = tex_format
         if tex_format == TextureFormat.RGB24:
             s_tex_format = TextureFormat.RGBA32
             pil_mode = "RGBA"
-        # elif tex_format == TextureFormat.BGR24:
-        #     s_tex_format = TextureFormat.BGRA32
+        elif tex_format == TextureFormat.BGR24:
```
isHarryh, line 232: BGR24 is not tested, but theoretically, there is no problem.

UnityPy/export/Texture2DConverter.py
```diff
@@ -219,50 +215,38 @@ def image_to_texture2d(
...
         enc_img = img.tobytes("raw", pil_mode)
 
-    return enc_img, tex_format
-
+    if platform == BuildTarget.Switch and platform_blob:
+        enc_img = TextureSwizzler.swizzle(
+            enc_img, width, height, texture_format, platform_blob
+        )
 
-def assert_rgba(img: Image.Image, target_texture_format: TextureFormat) -> Image.Image:
```
isHarryh, line 259: Never used function was removed.

K0lb3:
I implemented a fixed version before I saw that you included it as well, otherwise I would have sticked and requested a fair amount of changes.
Imo PRs that fix things should more or less stay atomar and only implement the fix, and nothing else.
This makes it easier to review and therefore gets merged faster.

Regarding the changes I would have requested, it would have started with using the formatting defined by rust.toml.
I might be at fault for that not being correct, as I borked the inclusion pattern in ruff.toml, but in the end it could still be seen in the git difference.

 Some asserts you included aren't really necessary, at least the type checker didn't complain to me about those points, and I don't really see how a type checker could get confused by some of them.
 
 Anyway, in case I didn't annoy you too much with my behavior,
 I would be happy to merge two separate PRs based on your changes.
 The alpha channel *fix* and the changes to etc and eac.

isHarryh:
Sorry for changing too many things in one single PR. I am also happy to draft new separated PRs later to fix these remaining things.

PR #324, opened 2025-06-01
Cleanup Texture2DConverter and fix typos in pyproject.toml

isHarryh:
### Summary

This PR does some minor changes to Texture2DConverter.

- ~~`platform_blob` can be `List[int]` in newer version of Unity, instead of only `bytes`.~~
- One `platform_blob` type hint still uses `bytes`, now corrected to `List[int]`.
- `if TYPE_CHECKING: assert xxx is not None` seems redundant, now removed `if`.
- ~~Pre-converts `platform` to `BuildTarget` enum.~~
- ~~Wraps swizzle texture_format conversion into TextureSwizzler.~~
- Simplifies the redundant if-statements in `image_to_texture2d`.
- Optimizes docstrings and typos.

Along with two typo fixes in the `pyproject.toml`:

- ~~Fixes the invalid license field in `pyproject.toml` which causes CodeQL failing.~~
- Fixes the typo of the keyword `data-minig` in `pyproject.toml`.

### DIFF
--- a/UnityPy/helpers/TextureSwizzler.py
+++ b/UnityPy/helpers/TextureSwizzler.py
@@ -1,5 +1,6 @@
 # based on https://github.com/nesrak1/AssetsTools.NET/blob/dev/AssetsTools.NET.Texture/Swizzle/SwitchSwizzle.cs
 from typing import Dict, List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 from ..enums import BuildTarget, TextureFormat
 
@@ -118,7 +119,9 @@ def get_switch_gobs_per_block(platform_blob: List[int]) -> int:
     return 1 << int.from_bytes(platform_blob[8:12], "little")
 
 
-def is_switch_swizzled(platform: Union[BuildTarget, int], platform_blob: Optional[List[int]]) -> bool:
+def is_switch_swizzled(
+    platform: Union[BuildTarget, int], platform_blob: Optional[List[int]]
+) -> bool:
     if platform != BuildTarget.Switch:
         return False
     if not platform_blob or len(platform_blob) < 12:
--- a/UnityPy/helpers/TextureSwizzler.py
+++ b/UnityPy/helpers/TextureSwizzler.py
@@ -70,7 +70,7 @@ def swizzle(
 
 
 # this should be the amount of pixels that can fit 16 bytes
-TEXTUREFORMAT_BLOCK_SIZE_MAP: Dict[TextureFormat, Tuple[int, int]] = {
+TEXTURE_FORMAT_BLOCK_SIZE_MAP: Dict[TextureFormat, Tuple[int, int]] = {
     TextureFormat.Alpha8: (16, 1),  # 1 byte per pixel
     TextureFormat.ARGB4444: (8, 1),  # 2 bytes per pixel
     TextureFormat.RGBA32: (4, 1),  # 4 bytes per pixel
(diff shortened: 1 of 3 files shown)


K0lb3 (commented):
Sorry that I'm only replying now.

I will be blunt, I don't really see the point of this PR.

The typo fixes, doc string adjustments and the BC format collapsing are okay, but that's it.

``TextureSwizzler.get_corrected_texture_format`` is basically a one-time use function,
which only adds a layer of indirection without any benefit, as at one of the two places where it's being used, the same check has to be done again anyway.

``TextureSwizzler.PlatformBlobType`` is also not really useful.
The usage of bytes from before was actually wrong, as the platformblob type is defined by the Texture2D field of it, so just List[int] is enough.

The platform type also doesn't have to be narrowed down to BuildTarget,
as the only use of platform is in an equality check, which works just fine with an int as well.

isHarryh:
This PR now contains:

1. Typo and docstring adjustments.
2. BC format if-statements collapsing.

with some new changes:

3. One `platform_blob` type hint still uses `bytes`, now corrected to `List[int]`.
4. `if TYPE_CHECKING: assert xxx is not None` seems redundant, now removed `if`.
5. Fixs the invalid license field in `pyproject.toml` which causes CodeQL failing.
6. Fixs the typo of the keyword `data-minig` in `pyproject.toml`.

pyproject.toml
```diff
@@ -7,14 +7,14 @@ name = "UnityPy"
 authors = [{ name = "Rudolf Kolbe", email = "rkolbe96@gmail.com" }]
 description = "A Unity extraction and patching package"
 readme = "README.md"
-license = "MIT"
+license = { text = "MIT" }
```
K0lb3, line 10: Directly writing a [SPDX license expression](https://packaging.python.org/en/latest/glossary/#term-License-Expression) is the new expected value for the license field, see [Writing your pyproject.toml - Python Packaging User Guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license).

UnityPy/helpers/TextureSwizzler.py
```diff
@@ -70,7 +70,7 @@ def swizzle(
 
 
 # this should be the amount of pixels that can fit 16 bytes
-TEXTUREFORMAT_BLOCK_SIZE_MAP: Dict[TextureFormat, Tuple[int, int]] = {
+TEXTURE_FORMAT_BLOCK_SIZE_MAP: Dict[TextureFormat, Tuple[int, int]] = {
```
K0lb3, line 73: This is actually a breaking change, as the export of ``TEXTUREFORMAT_BLOCK_SIZE_MAP`` isn't restricted.
Therefore the commit containing this change has to be marked according with a !.
As the commit message of e3a398c doesn't indicate that the rename was done in it,
it would be better to move this change into a separate commit.

pyproject.toml
K0lb3: Thanks for clarifying that.
I guess that I will simply have to set the min version to 3.9.

3.7 is already broken due to typing_extensions used in some dependency, so might as well remove 3.8 support as well.
As there are already breaking commits up, this can come with the next minor update.

isHarryh: Imo, we might consider delaying the restriction on the maximum version of setuptools, as some projects that currently rely on the latest version of setuptools may encounter dependency resolution errors.

K0lb3: I will simply set to to a possible future version, going by the current release cadence for now, as I would probably forget about it otherwise.

You have a pretty good point there that I didn't even think about.
This makes the whole issue very tricky.

I prefer to support old versions as long as it doesn't take too much effort to support them,
and there are certainly options to work around the issue, just that they are a bit too hacky imo.

K0lb3 (dismissed):

PR #331, opened 2025-07-09
Prepend monobehaviour base fields to generated typetrees

jakobhellermann:
Complementary PR to https://github.com/UnityPy-Org/TypeTreeGeneratorAPI/pull/4

This should make all backends usable, whereas previously only the `AssetStudio` backend would include monobehaviour base fields.

Now, the generator generates only the script fields, and `UnityPy` prepends the fields.

Needs a version bump of the generator when that is released.

### DIFF
--- a/UnityPy/helpers/TypeTreeGenerator.py
+++ b/UnityPy/helpers/TypeTreeGenerator.py
@@ -43,7 +43,7 @@ class TypeTreeGenerator(TypeTreeGeneratorBase):
                 data = f.read()
                 self.load_dll(data)
 
-    def get_nodes_up(self, assembly: str, fullname: str) -> TypeTreeNode:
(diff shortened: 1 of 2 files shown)


UnityPy/helpers/TypeTreeGenerator.py
```diff
@@ -60,6 +60,7 @@ def get_nodes_up(self, assembly: str, fullname: str) -> TypeTreeNode:
             0,
             0,
             m_MetaFlag=base_root.m_MetaFlag,
+            m_Children=base_node.m_Children[:],
```
K0lb3, line 63: Hard to say if a copy is really necessary.

K0lb3:
I will merge this PR today or tomorrow, after I push the generator update as new version, so that the changes here include the updated generator version as minimal dependency version.

PR #334, opened 2025-07-17
Fix ZDE in PackedBitVector

### LINKED ISSUE #333: PackedBitVector.unpack_floats: ZeroDivisionError
**Code**
```python
import UnityPy
from UnityPy.enums import ClassIDType
from UnityPy.classes import SkinnedMeshRenderer
from UnityPy.helpers.MeshHelper import MeshHandler

env = UnityPy.load(
    "assets-_mx-characters-shiroko_original-_mxdependency-2024-11-18_assets_all_1278508529.bundle"
)
for obj in filter(lambda o: o.type == ClassIDType.SkinnedMeshRenderer, env.objects):
    obj: SkinnedMeshRenderer = obj.read()
    mesh = obj.m_Mesh.read()
    print(mesh.m_Name, "enter")
    handler = MeshHandler(mesh)
    handler.process()
    print(mesh.m_Name, "ok")
    pass
```

**Error**
```bash
Shiroko_Dron enter
Shiroko_Dron ok
Shiroko_Dron_Wing enter
Traceback (most recent call last):
  File "C:\Users\mos9527\Desktop\Bundles\main.py", line 14, in <module>
    handler.process()
  File "C:\Users\mos9527\miniconda3\Lib\site-packages\UnityPy\helpers\MeshHelper.py", line 201, in process
    self.decompress_compressed_mesh()
  File "C:\Users\mos9527\miniconda3\Lib\site-packages\UnityPy\helpers\MeshHelper.py", line 543, in decompress_compressed_mesh
    normalData = unpack_floats(m_CompressedMesh.m_Normals, shape=(2,))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mos9527\miniconda3\Lib\site-packages\UnityPy\helpers\PackedBitVector.py", line 78, in unpack_floats
    scale = packed.m_Range / ((1 << packed.m_BitSize) - 1)
            ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ZeroDivisionError: float division by zero
```

**Bug**
Both `m_Range` and `m_BitSize` (!!) are read as `0` as shown in the debugger view attached below.
Image

**To Reproduce**
- a copy of the file that causes the problem

[assets-_mx-characters-shiroko_original-_mxdependency-2024-11-18_assets_all_1278508529.zip](https://github.com/user-attachments/files/21246609/assets-_mx-characters-shiroko_original-_mxdependency-2024-11-18_assets_all_1278508529.zip)

- following data:
  - Python version: 3.11.5
  - UnityPy version: 1.23.0

isHarryh:
### Summary

Fixes #333 

### Test

Tested with the sample file provided in #333 .

(Left: AssetStudio, Right: Ours)

image

### DIFF
--- a/UnityPy/helpers/PackedBitVector.py
+++ b/UnityPy/helpers/PackedBitVector.py
@@ -1,10 +1,10 @@
-from typing import TYPE_CHECKING, List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, List, Optional, Tuple
 
 if TYPE_CHECKING:
     from ..classes.generated import PackedBitVector
 
 
-def reshape(data: list, shape: Optional[Tuple[int, ...]] = None) -> list:
+def reshape(data: list, shape: Optional[Tuple[int, ...]] = None) -> List[Any]:
     if shape is None:
         return data
     if len(shape) == 1:
@@ -12,7 +12,7 @@ def reshape(data: list, shape: Optional[Tuple[int, ...]] = None) -> list:
         return [data[i : i + m] for i in range(0, len(data), m)]
     elif len(shape) == 2:
         m, n = shape
-        return [[[data[i + j : i + j + n] for j in range(0, m * n, n)]] for i in range(0, len(data), m * n)]
+        return [[data[i + j : i + j + n] for j in range(0, m * n, n)] for i in range(0, len(data), m * n)]
     else:
         raise ValueError("Invalid shape")
 
@@ -22,7 +22,7 @@ def unpack_ints(
     start: int = 0,
     count: Optional[int] = None,
     shape: Optional[Tuple[int, ...]] = None,
-) -> List[int]:
+) -> List[Any]:
     assert packed.m_BitSize is not None
 
     m_BitSize = packed.m_BitSize
@@ -70,7 +70,7 @@ def unpack_floats(
     start: int = 0,
     count: Optional[int] = None,
     shape: Optional[Tuple[int, ...]] = None,
-) -> List[float]:
+) -> List[Any]:
     assert packed.m_BitSize is not None and packed.m_Range is not None and packed.m_Start is not None
 
     # avoid zero division of scale
--- a/UnityPy/helpers/PackedBitVector.py
+++ b/UnityPy/helpers/PackedBitVector.py
@@ -73,10 +73,15 @@ def unpack_floats(
 ) -> List[float]:
     assert packed.m_BitSize is not None and packed.m_Range is not None and packed.m_Start is not None
 
-    # read as int and cast up to double to prevent loss of precision
-    quantized_f64 = unpack_ints(packed, start, count)
-    scale = packed.m_Range / ((1 << packed.m_BitSize) - 1)
-    quantized = [x * scale + packed.m_Start for x in quantized_f64]
+    # avoid zero division of scale
+    if packed.m_BitSize == 0:
+        quantized = [packed.m_Start] * (packed.m_NumItems if count is None else count)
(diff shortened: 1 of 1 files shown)


UnityPy/helpers/PackedBitVector.py
```diff
@@ -75,7 +75,14 @@ def unpack_floats(
 
     # read as int and cast up to double to prevent loss of precision
     quantized_f64 = unpack_ints(packed, start, count)
-    scale = packed.m_Range / ((1 << packed.m_BitSize) - 1)
+
+    # avoid zero division
+    scale_denominator = (1 << packed.m_BitSize) - 1
+    if scale_denominator == 0:
```
K0lb3, line 81: In the original C++ code we have

```cpp
uint32_t x;
float scale;

x &= (1 << m_BitSize) - 1;
data[i] = (x / (scale * ((1 << (m_BitSize)) - 1))) + m_Start;
```

Assuming that m_BitSize is 0, then x is 0 as well.
With this we get

```cpp
data[i] = (0 / (scale * 0)) + m_Start;
```

Based on the IEEE 754 rules:

finite * 0.0 = 0.0 → 0 / 0.0 = NaN
±∞ * 0.0 = NaN → 0 / NaN = NaN
NaN * 0.0 = NaN → 0 / NaN = NaN

so we the result will always be NaN.

Which means that we can effectively return early with an array of NaN.

K0lb3 (changes requested):
Thanks, there is just a little extra optimization I would like for this special case.

isHarryh:
I agree with @mos9527 , we probably should return an array of `m_Start` here.

UnityPy/helpers/PackedBitVector.py
```diff
@@ -22,7 +22,7 @@ def unpack_ints(
     start: int = 0,
     count: Optional[int] = None,
     shape: Optional[Tuple[int, ...]] = None,
-) -> List[int]:
+) -> List[Any]:
```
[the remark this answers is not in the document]
K0lb3: No need for the trouble, as this part is on the chopping block (rewrite on a comming minor or major update) anyway, so List[Any] is enough

K0lb3:
Sorry for my lack of reply.

I was thinking about the issue today and talked it from a different perspective, how a bitsize of 0 could occur.
A bitsize of 0 can only occur if the max and min value of the are identical, so if the array only has identical elements.
In that case all elements would be 0 after the smalled element gets substracted as offset from each element.
This offset is stored as m_Start.

So it certainly makes sense to simply assume an array of ``[m_Start]*m_NumItems`` as decoded result (before reshaping).
You probably had similar ideas as well.

UnityPy/helpers/PackedBitVector.py
```diff
@@ -70,13 +70,19 @@ def unpack_floats(
...
     assert packed.m_BitSize is not None and packed.m_Range is not None and packed.m_Start is not None
 
     # read as int and cast up to double to prevent loss of precision
     quantized_f64 = unpack_ints(packed, start, count)
-    scale = packed.m_Range / ((1 << packed.m_BitSize) - 1)
-    quantized = [x * scale + packed.m_Start for x in quantized_f64]
+
+    # avoid zero division of scale
+    if packed.m_BitSize == 0:
```
K0lb3, line 80: The bitsize check can be moved to before the int unpacking.

K0lb3 (approved):

PR #335, opened 2025-07-20
Refactor MeshHelper

isHarryh:
### Summary

This PR does some improvements to MeshHelper.

1. Making `flat_list_to_tuples` an inline function.
2. Introduces the `lists_to_tuples` function to fix the type error that `List[List]` produced by the `zeros` function cannot be assigned to tuple-member-lists such as `List[Tuple3f]` `List[Tuple4f]`.
3. Refactors the `get_triangles` function.
   1. Changes the way to generate triangles lists.
   2. Adds a `indexCount` fix for TriangleStrip submeshes, see [AssetStudio's code](https://github.com/aelurum/AssetStudio/blob/7b7eac62d80c8ef9a596cac7df727a2c81044a9f/AssetStudio/Classes/Mesh.cs#L1179).
   3. Removes the TODOs in the `get_triangles` function (I have tested that using NumPy `as_strided` function cannot realize a speed-up).

### DIFF
--- a/UnityPy/helpers/MeshHelper.py
+++ b/UnityPy/helpers/MeshHelper.py
@@ -625,10 +625,10 @@ class MeshHandler:
 
     def get_triangles(self) -> List[List[Tuple[int, ...]]]:
         assert self.m_IndexBuffer is not None
+        assert self.src.m_SubMeshes is not None
 
         submeshes: List[List[Tuple[int, ...]]] = []
 
-        assert self.src and self.src.m_SubMeshes is not None, "No submesh data!"
         for m_SubMesh in self.src.m_SubMeshes:
             firstIndex = m_SubMesh.firstByte // 2
             if not self.m_Use16BitIndices:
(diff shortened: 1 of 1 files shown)


UnityPy/helpers/MeshHelper.py
```diff
@@ -650,41 +654,31 @@ def get_triangles(self) -> List[List[Tuple[int, ...]]]:
...
-                triangles = self.m_IndexBuffer[firstIndex : firstIndex + indexCount]  # type: ignore
-                triangles = [triangles[i : i + 3] for i in range(0, len(triangles), 3)]  # type: ignore
-            elif self.version[0] < 4 or topology == MeshTopology.TriangleStrip:  # TriangleStrip
-                # todo: use as_strided, then fix winding, finally remove degenerates
-                triIndex = 0
-                triangles = [None] * (indexCount - 2)  # type: ignore
-
-                for i in range(indexCount - 2):
-                    a, b, c = self.m_IndexBuffer[firstIndex + i : firstIndex + i + 3]
+                for i in range(firstIndex, firstIndex + indexCount, 3):
```
K0lb3, line 660: Python for loops are way slower than list comprehension.
While it's tricky to use it for the other versions,
it's rather easy for normal triangles to use, and should therefore be used.

UnityPy/helpers/MeshHelper.py
```diff
@@ -650,41 +654,31 @@ def get_triangles(self) -> List[List[Tuple[int, ...]]]:
             indexCount = m_SubMesh.indexCount
             topology = m_SubMesh.topology
 
-            triangles: List[Tuple[int, ...]]
+            triangles: List[Tuple[int, ...]] = []
```
K0lb3, line 657: There is no point is already assigning a value here, as we don't know the size yet.
Initializing the list with the correct size and writing into it is always faster than appending, as the later requires at least one, usually multiple reallocs, which are expensive.

K0lb3 (changes requested):
see comments

K0lb3 (dismissed):

PR #338, opened 2025-07-30
MeshHandler: Fix `weightsData` dequantization error

mos9527:
In the Editor's serialization code, normalized weights are quantized as 5-bit fixed point UNORM values then packed as ints.

 This implies flooring - while the sum of the quantized values are guaranteed to be `31` this can lead to a loss of precision where the *de*quantized sum (div by 31 then sum) of weights is not correct with IEEE float.

Such cases can be trivally reproduced by search, for example a tuple of $$(1,25,5)$$ summed as $$(\frac{1}{31}+\frac{25}{31}+\frac{5}{31}$$ in _that exact order_ will yield $$0.9999999999999999$$ instead of $$1.0$$, and `((1/31+25/31+5/31) >= 1.0)` in Python interperter will evalutate as `False`.

This PR implements the sum of weight with ints instead. Alternatively an epsilon value could be chosen, but this is what the player code does.

Attachments in https://github.com/K0lb3/UnityPy/issues/333 can reproduce this issue. See screenshots below.

- Before patch:
image

- After patch:
image

### DIFF
--- a/UnityPy/helpers/MeshHelper.py
+++ b/UnityPy/helpers/MeshHelper.py
@@ -583,7 +583,6 @@ class MeshHandler:
         # Skin
         if m_CompressedMesh.m_Weights.m_NumItems > 0:
             weightsData = unpack_ints(m_CompressedMesh.m_Weights)
-            weightsData = [weight for weight in weightsData]
             boneIndicesData = unpack_ints(m_CompressedMesh.m_BoneIndices)
 
             vertexIndex = 0


UnityPy/helpers/MeshHelper.py
```diff
@@ -583,7 +583,7 @@ def decompress_compressed_mesh(self):
         # Skin
         if m_CompressedMesh.m_Weights.m_NumItems > 0:
             weightsData = unpack_ints(m_CompressedMesh.m_Weights)
-            weightsData = [weight / 31 for weight in weightsData]
+            weightsData = [weight for weight in weightsData]
```
K0lb3, line 586: This line could be removed now.

K0lb3:
Thanks for the write-up and the example.
Looks like I failed acting smart when I moved the division into an earlier list comprehension.
I totally missed the float accuracy issue when I did so.

K0lb3 (changes requested):
Just the removal of the mentioned line, and then it can be merged.

PR #342, opened 2025-08-22
feat: Lock-free ASTC codec operations

mos9527:
This PR removes the explict locking in ASTC related codec routines in favor of per-thread allocation instead.

For ST workloads this changes nothing - contexts will still be cached and is cheap to reuse. 

For MT the extra allocation will grow with calling threads - though may be reused as per [get_ident() docs](https://docs.python.org/3/library/threading.html#threading.get_ident). `functools.lru_cache` is used to mitigate potential leaks here.

This applies to both encoding and decoding processes.

## Addenum: Benchmark

- Code
```python

def test_encode(size):
    from UnityPy.export.Texture2DConverter import compress_astc, TextureFormat
    img = bytes(4 * size * size)
    compress_astc(img, size, size, TextureFormat.ASTC_RGB_6x6)

def test(executor, size, jobs, workers):
    for res in executor.map(test_encode, [size] * jobs):
        pass # Gather all results before returning

if __name__ == "__main__":
    t_repeat = 8
    t_number = 16
    size = 512
    jobs = 128
    max_workers = 6

    import timeit
    from concurrent.futures import ThreadPoolExecutor
    for workers in range(1, max_workers + 1):
        with ThreadPoolExecutor(max_workers=workers) as executor:
            times = timeit.repeat("test(executor, size, jobs, workers)", globals=globals(), number=t_number, repeat=t_repeat)     
            print(f'ThreadPool n={t_number} workers={workers} avg: {sum(times) / len(times)}')
  ```
  
  - Results (https://github.com/K0lb3/UnityPy/commit/a52f70d1020ba0e1b5325626070b433c31c8d22a)
  ```bash
ThreadPool n=16 workers=1 avg: 1.6013365874994179
ThreadPool n=16 workers=2 avg: 1.8349196000001484
ThreadPool n=16 workers=3 avg: 1.7442084625001826
ThreadPool n=16 workers=4 avg: 1.7662892000000738
ThreadPool n=16 workers=5 avg: 1.7565655249998144
ThreadPool n=16 workers=6 avg: 1.7495201125002495
```
- Results (This PR)
```bash
ThreadPool n=16 workers=1 avg: 1.7571594624998852
ThreadPool n=16 workers=2 avg: 0.9583396875002563
ThreadPool n=16 workers=3 avg: 0.6424626125001396
ThreadPool n=16 workers=4 avg: 0.514299999999821
ThreadPool n=16 workers=5 avg: 0.44599136250030824
ThreadPool n=16 workers=6 avg: 0.43532513749960344
```

### DIFF
--- a/UnityPy/export/Texture2DConverter.py
+++ b/UnityPy/export/Texture2DConverter.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import struct
 from io import BytesIO
-from threading import Lock
+from functools import lru_cache
+from threading import get_ident
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
 
 import astc_encoder
@@ -108,9 +109,8 @@ def compress_astc(data: bytes, width: int, height: int, target_texture_format: T
     assert block_size is not None, f"failed to get block size for {target_texture_format.name}"
     swizzle = astc_encoder.ASTCSwizzle.from_str("RGBA")
 
-    context, lock = get_astc_context(block_size)
-    with lock:
-        enc_img = context.compress(astc_image, swizzle)
+    context = get_astc_context(block_size)
+    enc_img = context.compress(astc_image, swizzle)
 
     return enc_img
 
@@ -359,32 +359,31 @@ def astc(image_data: bytes, width: int, height: int, block_size: tuple) -> Image
     if len(image_data) < texture_size:
         raise ValueError(f"Invalid ASTC data size: {len(image_data)} < {texture_size}")
 
-    context, lock = get_astc_context(block_size)
-    with lock:
-        context.decompress(image_data[:texture_size], image, astc_encoder.ASTCSwizzle.from_str("RGBA"))
+    context = get_astc_context(block_size)
+    context.decompress(image_data[:texture_size], image, astc_encoder.ASTCSwizzle.from_str("RGBA"))
     assert image.data is not None, "Decompression failed, image data is None"
 
     return Image.frombytes("RGBA", (width, height), image.data, "raw", "RGBA")
 
 
-ASTC_CONTEXTS: Dict[Tuple[int, int], Tuple[astc_encoder.ASTCContext, Lock]] = {}
-
+@lru_cache(maxsize=128)
+def _get_astc_context(ident: int, block_size: tuple):
+    config = astc_encoder.ASTCConfig(
+        astc_encoder.ASTCProfile.LDR,
+        *block_size,
+        block_z=1,
+        quality=100,
+        flags=astc_encoder.ASTCConfigFlags.USE_DECODE_UNORM8,
+    )
+    context = astc_encoder.ASTCContext(config)
+    return context    
 
 def get_astc_context(block_size: tuple):
-    """Get the ASTC context and its lock using the given `block_size`."""
-    if block_size not in ASTC_CONTEXTS:
-        config = astc_encoder.ASTCConfig(
-            astc_encoder.ASTCProfile.LDR,
-            *block_size,
-            block_z=1,
-            quality=100,
-            flags=astc_encoder.ASTCConfigFlags.USE_DECODE_UNORM8,
-        )
-        context = astc_encoder.ASTCContext(config)
-        lock = Lock()
-        ASTC_CONTEXTS[block_size] = (context, lock)
-    return ASTC_CONTEXTS[block_size]
-
+    """Get the ASTC context for the current thread using the given `block_size`.
+    Created contexts belong to and only to the calling thread, and may be cached.
+    This function is thread safe.
+    """
+    return _get_astc_context(get_ident(), block_size)
 
 def calculate_astc_compressed_size(width: int, height: int, block_size: tuple) -> int:
     """Calculate the size of the compressed data for ASTC."""


isHarryh:
I tested the memory performance of this PR using `tracemalloc`. In the worst-case scenario from my test, the `context` cache occupies ~240KB of extra memory, which is **acceptable** for most users.

Btw, I would like to reduce the `maxsize` of `lru_cache` to `64`, because as you can see in `CONV_TABLE`, there are only a few different types of ASTC block sizes.

isHarryh (approved):

isHarryh:
> Unfortunately the memory situation is infact quite unpleasant. `tracemalloc` only tracks memory allocation made by Python's allocators, which upstream `astc-encoder` does not use, meaning actual codec alloc is unaccounted for.

After correcting my earlier mistake, I now use `psutil` to profile memory usage instead of `tracemalloc`, and it confirms that the memory footprint of this PR’s solution is **significant**.

Each context cache consumes \~3.5 MB, so 128 contexts result in \~400 MB memory. This means the new solution may negatively affect **multi-threaded users** who are unaware of the change.

Fortunately:

- The memory usage **scales proportionally** with the `maxsize` of  `lru_cache`.
- Stale context **can be freed correctly** by `lru_cache` (tested with `astc_encoder` v0.1.11).

### Test

Code:

<details>

```py
import queue
import timeit
import psutil
from concurrent.futures import ThreadPoolExecutor
from PIL import Image
from UnityPy.export.Texture2DConverter import compress_astc, astc, TextureFormat, _get_astc_context

SIZE = 1024
IMAGE_RAW = Image.new("RGBA", (SIZE, SIZE)).tobytes("raw", "RGBA")
IMAGE_ENCODED = compress_astc(IMAGE_RAW, SIZE, SIZE, TextureFormat.ASTC_RGB_4x4)


def encode():
    compress_astc(IMAGE_RAW, SIZE, SIZE, TextureFormat.ASTC_RGB_4x4)


def decode():
    astc(IMAGE_ENCODED, SIZE, SIZE, (4, 4))


def multi_threading(target, tasks, threads):
    task_queue = queue.Queue()
    for _ in range(tasks):
        task_queue.put(None)

    def worker():
        while True:
            try:
                task_queue.get(block=False)
                target()
                task_queue.task_done()
            except queue.Empty:
                break

    with ThreadPoolExecutor(max_workers=threads) as executor:
        futures = [executor.submit(worker) for _ in range(threads)]
        for future in futures:
            future.result()


if __name__ == "__main__":
    tasks = 10
    repeat = 10
    mem0 = psutil.Process().memory_info().rss

    for target in [encode, decode]:
        for threads in [1, 2, 4, 8]:
            time = timeit.timeit(lambda: multi_threading(target, tasks, threads), number=repeat)
            print(f"{target.__name__}: threads={threads}, speed={time / repeat:.6f}s / {tasks}it")

    mem1 = psutil.Process().memory_info().rss
    print(f"memory: current={mem1 >> 20}MB diff={(mem1 - mem0) >> 20}MB")

    _get_astc_context.cache_clear()

    mem2 = psutil.Process().memory_info().rss
    print(f"memory: current={mem2 >> 20}MB diff={(mem2 - mem1) >> 20}MB")
```

</details>

Output:

```txt
encode: threads=1, speed=0.158524s / 10it
encode: threads=2, speed=0.082843s / 10it
encode: threads=4, speed=0.052776s / 10it
encode: threads=8, speed=0.041788s / 10it
decode: threads=1, speed=0.141386s / 10it
decode: threads=2, speed=0.074519s / 10it
decode: threads=4, speed=0.049538s / 10it
decode: threads=8, speed=0.041709s / 10it
memory: current=551MB diff=438MB
memory: current=108MB diff=-443MB
```

### Suggestions

1. Use the lock-free solution only, but set `maxsize` to a very small value (e.g., `4` or `8`):
   - ST users: Context regeneration only occurs if many block sizes are used, which is rare.
   - MT users: Context regeneration happens more often, but they take only \~10% of total ASTC (de)compression time - negligible compared with the \~4x MT speedup from this PR.
2. Expose `_get_astc_context.cache_clear()` API to allow user to free cache manually.

K0lb3:
Pillow sadly only really support RGBA at 8 bit.
HDR images would be ripped down to it in Pillow.

Supporting HDR would require supporting other "interchange" formats/libs than Pillow.
I have this planned, but welp, time and interest tend to be lacking -.-'

I guess that I could let my c-extensions use Py_malloc and co., but that still wouldn't change anything about the underlying libs, besides if they would be compiled with no-std and then linking it against a modified std which maps malloc and co to the python ones.

K0lb3:
Sorry for the delay,
I will try to review it for good tomorrow or at the start of next week.

As I want to move the fmod libs out into a separate dependency to reduce the binary version size of UnityPy,
the next release version is depending on the progress on that front....which got some weird issues that don't happen at UnityPy despite basically identical code -.-'

K0lb3:
Going by the docs and code for lru_cache, it doesn't create a cache for each thread.
It simply handles the parallel access to a cache and makes that thread-safe.

By using it, the actual problem of using the same context in parallel for different (de)compression jobs can occur.

mos9527:
> Going by the docs and code for lru_cache, it doesn't create a cache for each thread. It simply handles the parallel access to a cache and makes that thread-safe.
> 
> By using it, the actual problem of using the same context in parallel for different (de)compression jobs can occur.

It shouldn't. As context cache is key'd by the calling thread's ident.

https://github.com/mos9527/UnityPy/blob/6052162e048fd606f77883414d0b2bd52963476d/UnityPy/export/Texture2DConverter.py#L369-L386

In hindsight, having a *global* cache, or a *global* state in a *library* at all is a pretty awful idea. If caching the contexts is really that much needed (Question: how 'slow' would allocating and then deallocating a context -per encode/decode call-   be?), perhaps tying its lifetime to `Environment` is a better idea. Though the memory usage is still unpredictable and might be an issue for users with constrained resources.

K0lb3:
Oh, sorry, I totally missed the ident.

The context creation is rather expensive as far as I recall, so caching it makes sense imo.
I guess that the better solution might be to simply unlock the parallel processing of the de/comp function,
which should offset the multi-threading lock.

K0lb3:
For now I would say that constrained resources shouldn't be seen as a problem yet,
as other things like simply big compressed bundles can consume way more memory.

Systems with a low memory ceiling likely can't support many threads anyway,
so the cost of having multiple contexts cached shouldn't be that high for them.

mos9527:
> On that note, lru_cache isn't even needed imo, in fact, it might cause a slight slowdown due to sorting by recently used items. Simply cache or lru_cache(0) should be a slight bit better without costing more memory for real, as the cache size likely won't be filled anyway.

In `functools` impl LRU order is [maintained with a circular linked list](https://github.com/python/cpython/blob/056d6c5ed90bfed2861098f1e42640d6ea62cac8/Lib/functools.py#L668-L711), which is $O(1)$ for the cache update itself. The *real* slowdown could come from the destruction of the older contexts, though again:

> The size is quite arbitrary - though again it's not really there to relieve memory usage. It's a very desperate measure against leaks from e.g. new worker threads not reusing idents or way too much workers in general. 

TL;DR - LRU is here to prevent leaking. A size-limited cache would be a necessary evil to *accommodate* misuse (e.g. stated as above) since the current API do allow it.

PR #345, opened 2025-09-14
Update BundleFile.py

AniccaY:
Fix CN bundle decrypt :
The wrong block was decrypted
only LZ4HC need decrypt

### DIFF
--- a/UnityPy/files/BundleFile.py
+++ b/UnityPy/files/BundleFile.py
@@ -489,7 +489,7 @@ class BundleFile(File.File):
         """
         Parameters
         ----------
-        raw_compressed_data : bytes
+        compressed_data : bytes
             The compressed data.
         uncompressed_size : int
             The uncompressed size of the data.


AniccaY:
old version will decrypt type: None Block, cause read bundle error

isHarryh:
I'm interested in the asset file you try to decrypt. Could you please provide a sample file and its key to let me investigate deeper?

AniccaY:
[Uploading sample.zip…]()

isHarryh (changes requested):
I don't understand the significance of your new changes.

Please edit the if-statement of L505 only.

K0lb3:
I can only second all of @isHarryh comments.

I did some digging through the TuanjiePlayer just now,
and as far as I can see, there is only an early return path for no compression.
Every other compression type can be encrypted.

K0lb3:
As the PR had unwanted changes and was based on "guesswork" based on the work of others,
while I checked the actual engine code, I opted to simply close this PR and push the fix directly myself.

PR #357, opened 2026-01-24
[TESTS] Add global random seed for deterministic test

Stephen0512:
## Description

This PR sets a fixed global random seed after imports in the test file to ensure deterministic and reproducible test behavior.

## Motivation

Some tests rely on randomness without a fixed seed, which can lead to flaky tests, inconsistent CI results, and difficulty reproducing failures.

## Change

A global random seed is initialized to make random behavior consistent across test runs.

## Benefits

- More reproducible and reliable tests  
- Easier debugging of failures  
- Consistent CI behavior  

The issue was identified during an ongoing research project.

### DIFF
--- a/tests/test_typetree.py
+++ b/tests/test_typetree.py
@@ -13,6 +13,7 @@ from UnityPy.helpers.TypeTreeNode import TypeTreeNode
 from UnityPy.helpers.UnityVersion import UnityVersion
 from UnityPy.streams import EndianBinaryReader, EndianBinaryWriter
 
+random.seed(17)
 PROCESS = psutil.Process(os.getpid())
 
 


isHarryh:
@K0lb3 **I suggest to reject this PR.**

## Reasons

1. This author have made numerous similar PR that has the same pattern, like "Add global random seed for deterministic test" and "Explicitly close filesto preventresource leak". I believe the intention behind his action is questionable. df507e1281b94fa7543934c2c763c597

2. This PR doesn't cover all the tests - it just adds fixed random seed to *one* of the four tests.

3. This test has no randomness in fact. **No need to add fixed random seed.**

PR #372, opened 2026-05-18
feat: read preload table to get path id

StarHeartHunt:
[battle_enm_pfb_18.ab.zip](https://github.com/user-attachments/files/27982814/battle_enm_pfb_18.ab.zip)

for some assets in this assetbundle, container paths are defined in preload table

reference: https://github.com/Perfare/AssetStudio/blob/master/AssetStudioGUI/Studio.cs#L210

### DIFF
--- a/UnityPy/environment.py
+++ b/UnityPy/environment.py
@@ -206,7 +206,9 @@ class Environment:
         return search(self)
 
     def _build_container_index(self) -> None:
-        self._container_index_built = True
+        if not self._container_index_built:
+            self._container_index_built = True
+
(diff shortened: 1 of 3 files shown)


K0lb3:
Thanks,
and sorry for the late response.

The PR looks good, except one little detail.
Currently the ``_build_container_index`` gets called all the time when accessing a container.
Doing so once when creating the ``ContainerHelper`` is a better solution.

K0lb3 (changes requested):
see previous comment

K0lb3 (approved):

PR #373, opened 2026-05-22
Checking for python version for the open_binary and files calls

TedyonGit:
I just wanted to fix the warning "DeprecationWarning".
So i just added a check for python version. If the python version is >= 3.9, then is going to use files call.
If not, is going to fallback to open_binary.

Is not much, but i hope it helps :D

### DIFF
--- a/UnityPy/helpers/Tpk.py
+++ b/UnityPy/helpers/Tpk.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import sys
 from enum import IntEnum, IntFlag
 from io import BytesIO
 from struct import Struct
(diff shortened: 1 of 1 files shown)


K0lb3:
Thanks,
and sorry for the late response.

I think that there was a reason I didn't implement this, but I can't recall it anymore.
Maybe I also simply did something wrong at that time, so let's see if the tests pass across different python versions and systems.

UnityPy/helpers/Tpk.py
```diff
@@ -18,8 +17,19 @@
 
 
 def init():
-    with open_binary("UnityPy.resources", "lzma.tpk") as f:
-        data = f.read()
+    try:
```
K0lb3: Instead of try/except,
``sys.version_info >= (3, 9)``
would be a better option

K0lb3 (changes requested):

K0lb3:
^ tests failed due to a ruff formatting check

K0lb3 (approved):

PR #374, opened 2026-05-30
Fix segfault in TypeTreeHelper boost when reading past buffer end

SNWCreations:
### Problem

When `read_typetree` (C++ boost path) parses certain MonoBehaviour objects whose type tree describes more data than is actually available in the object's byte range, the parser crashes with an access violation (exit code `0xC0000005` on Windows) instead of raising a Python exception.

This happens with MonoBehaviours that have complex serialization structures (e.g. `[SerializeReference]` fields with `ManagedReferencesRegistry`). The embedded type tree correctly describes the field structure, but the actual serialized data for some fields is shorter than what the type tree implies. Asset Studio handles the same scenario gracefully with a warning:

```
[Warning] Error reading field 'references': Unable to read beyond the end of the stream.
[Info] Error while read type, read 16776 bytes but expected 18320 bytes
```

UnityPy's pure Python `read_value` path would also handle this correctly (bounds checks raise Python exceptions). The crash only occurs in the C++ boost path.

### Root Cause

When a bounds check fires deep in a recursive `read_typetree_value` call chain, the function correctly returns `NULL` and sets `PyErr`. However, if the error occurs at a point where the caller has already partially consumed data and subsequent recursive calls are made before the NULL propagates fully, the parser can enter an inconsistent state. Additionally, `PyList_New(length)` calls were not checked for NULL — if a large (but positive) length value is read from misaligned data, the allocation can fail and subsequent `PyList_SET_ITEM` on the NULL pointer causes the segfault.

### Fix

1. **Added `exhausted` flag to `ReaderT`**: Once any bounds check fails, this flag is set. At the entry of `read_typetree_value` and `read_typetree_value_array`, the flag is checked and NULL is returned immediately. This ensures that even if one code path doesn't perfectly propagate NULL, all subsequent parsing attempts fail fast with a proper Python `ValueError`.

2. **Added NULL checks after `PyList_New`**: In the generic array path and `read_pair_array`, the return value of `PyList_New` is now checked before use.

3. **Set `exhausted` flag in all bounds-check failure paths**: Every `read_*` function that detects an out-of-bounds condition now sets `reader->exhausted = true` before returning NULL/error.

### Impact

- Callers of `obj.read()` will now receive a `ValueError` exception instead of a process crash when the object data cannot be fully parsed
- No change in behavior for objects that parse successfully
- The fix is defensive — it doesn't attempt to partially parse or recover data, it simply ensures the failure mode is a catchable Python exception rather than a segfault

### Reproduction

Any Unity bundle containing MonoBehaviour objects with `[SerializeReference]` fields where the serialized data is shorter than the type tree's full structure. The crash occurs when calling `obj.read()` on such objects with UnityPyBoost available.

Sample files:
[erroring_bundles.zip](https://github.com/user-attachments/files/28416913/erroring_bundles.zip)

### DIFF
--- a/UnityPyBoost/TypeTreeHelper.cpp
+++ b/UnityPyBoost/TypeTreeHelper.cpp
@@ -15,7 +15,6 @@ typedef struct Reader
     uint8_t *ptr;
     uint8_t *end;
     uint8_t *start;
-    bool exhausted;
 } ReaderT;
 
 typedef struct TypeTreeReaderConfig
@@ -76,7 +75,6 @@ inline PyObject *read_bool(ReaderT *reader)
 {
     if (reader->ptr + 1 > reader->end)
     {
-        reader->exhausted = true;
         PyErr_SetString(PyExc_ValueError, "read_bool out of bounds");
         return NULL;
     }
@@ -89,7 +87,6 @@ inline PyObject *read_bool_array(ReaderT *reader, int32_t count)
 {
     if (reader->ptr + count > reader->end)
     {
-        reader->exhausted = true;
         PyErr_SetString(PyExc_ValueError, "read_bool_array out of bounds");
         return NULL;
     }
@@ -107,7 +104,6 @@ inline PyObject *read_u8(ReaderT *reader)
 {
     if (reader->ptr + 1 > reader->end)
     {
-        reader->exhausted = true;
         PyErr_SetString(PyExc_ValueError, "read_u8 out of bounds");
         return NULL;
     }
@@ -118,7 +114,6 @@ inline PyObject *read_u8_array(ReaderT *reader, int32_t count)
 {
     if (reader->ptr + count > reader->end)
     {
-        reader->exhausted = true;
         PyErr_SetString(PyExc_ValueError, "read_u8_array out of bounds");
         return NULL;
     }
@@ -134,7 +129,6 @@ inline PyObject *read_s8(ReaderT *reader)
 {
     if (reader->ptr + 1 > reader->end)
     {
-        reader->exhausted = true;
         PyErr_SetString(PyExc_ValueError, "read_s8 out of bounds");
         return NULL;
     }
@@ -145,7 +139,6 @@ inline PyObject *read_s8_array(ReaderT *reader, int32_t count)
 {
     if (reader->ptr + count > reader->end)
     {
-        reader->exhausted = true;
         PyErr_SetString(PyExc_ValueError, "read_s8_array out of bounds");
         return NULL;
     }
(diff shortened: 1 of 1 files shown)


nesrak1:
Unrelated to the PR which would be good to have, your sample files look fine so this is just a bug in parsing SerializeReference in whatever version of AssetStudio you're using and UnityPy. It may be a case of nested managedReference or some other bug. We should probably make a new Github issue about it.

SNWCreations:
> Unrelated to the PR which would be good to have, your sample files look fine so this is just a bug in parsing SerializeReference in whatever version of AssetStudio you're using and UnityPy. It may be a case of nested managedReference or some other bug. We should probably make a new Github issue about it.

Thanks for the note. Just to clarify, this PR addresses exactly the crash scenario described: when the C++ boost path encounters a SerializeReference/ManagedReferencesRegistry structure where the serialized data is shorter than the type tree implies, it currently segfaults instead of raising a Python exception. The fix ensures it fails gracefully with a ValueError. Whether the data mismatch itself is a Unity serialization quirk or an upstream bug is a separate question, but at least UnityPy won't crash on it anymore.

K0lb3 (commented):
I'm not sure if the exhausted is really necessary,
given an exhaustive usage of PyErr_SetString and checking for NULL returns.

I think that the reason for the segfault is that I used ``PyErr_Format`` as return for ``read_num(_array)``, which breaks the return check.

K0lb3:
Also thanks for the PR,
and sorry for my late reply.

I will look into it more during this weekend.

SNWCreations:
Thanks for the review! You're right that `read_num`/`read_num_array` using `return PyErr_Format(...)` is likely the main culprit - `PyErr_Format` does return NULL, so the return value itself is correct, but if fixing that path alone resolves the segfault for you, I'm happy to simplify the PR to just that fix plus the `PyList_New` NULL checks.

The `exhausted` flag was added as a defensive measure: once any bounds check fails, subsequent recursive calls into `read_typetree_value` bail out immediately at the entry point rather than relying on every call site to perfectly propagate NULL. It trades a bit of added code for a stronger invariant. But I understand if you prefer to keep the fix minimal.

Happy to revise the PR in either direction - let me know your preference after you have had a look this weekend.

K0lb3:
Interestingly, when I tried to check it just now, ``parse_as_dict`` worked just fine, but ``parse_as_object`` crashed.
So the issue should be the class handling.

Never the less, the point of your PR still stands anyway, and actually has to be extended a bit.
All ``PyErr_Format`` returns or assignments have to be replaced with ``PyErr_SetString`` and a ``nullptr`` return.
Given that, a check for ``exhausted`` wouldn't be necessary.
Depending on the place, I'm all in for redundancy as safety layer, but imo this isn't such a place,
as the parsing is a hotloop and the safety layer might hide potential other errors that would have to be fixed.

SNWCreations:
For `read_num`/`read_num_array`, I kept `PyErr_Format` (call-then-return pattern) rather than switching to `PyErr_SetString`, since the format string includes `typeid(T).name()` to identify which type caused the bounds failure. The problematic pattern — using the return value of `PyErr_Format` directly as the function's return value — is now fixed. Let me know if you'd prefer a fixed string instead.

K0lb3 (approved):

K0lb3:
Sorry for the delay,
I got caught up in some stuff irl.

SNWCreations:
nvm, good to know the pr got merged.

